PR #7, opened 2020-02-17
Allow compile on Linux

KottV:


awwbees:
ah, excellent. I have plans in the very near future to set up a linux VM for testing bespoke, when I get that done I'll review this and merge it.

PR #34, opened 2020-08-23
Sampler: fix crash when pressing stop button with no sample

djdeath:


awwbees:
nice and simple. thanks!

PR #36, opened 2020-08-24
SamplePlayer: Add a button to load file from disk

djdeath:
Quite useful for MP3s etc...


awwbees:
nice, thanks! I made a few tweaks to adjust the layout and naming, but it's good to have a button in there. I also added a little bit of text to indicate that you can drag-and-drop samples from your file system right onto a sampleplayer, since that's not necessarily obvious (and maybe doesn't work in linux? I haven't tested that)

PR #38, opened 2020-08-24
RadioButton: Fix crash when no elements are added

djdeath:
Another crasher I found while trying pretty much all the components ;)

### DIFF
--- a/Source/RadioButton.cpp
+++ b/Source/RadioButton.cpp
@@ -203,6 +203,9 @@ ofVec2f RadioButton::GetOptionPosition(int optionIndex)
 
 void RadioButton::SetIndex(int i)
 {
+   if (mElements.empty())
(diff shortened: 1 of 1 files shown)


awwbees (changes requested):
if I understand the issue correctly, I think this would be better expressed as this, right?
```
if (mElements.size() == 0)
   return;
```

djdeath:
Most likely, I thought I could be covering potential other errors with `i >= mElements.size()`

awwbees:
ah thanks for making that change, this looks good! cases where a higher index is requested will avoid the crash via the ofClamp() call below, and it's possible that I'm relying on this behavior to always choose the last element of the list when you request a number out of range (instead of the radiobutton just failing to respond to the request)

PR #40, opened 2020-08-24
Gcc address sanitizer warnings

djdeath:
Trivial fixes reported by Gcc's version of address sanitizer.


awwbees:
nice! thank you for these.

PR #43, opened 2020-08-26
Add Roland TB-3 json description

djdeath:


awwbees:
thanks! always good to have more controllers mapped. I don't have a TB-3 so I can't test it, so I'll take your word for it.

PR #45, opened 2020-08-29
DrumPlayer: fix asan heap-buffer-overflow error

djdeath:


awwbees:
ah! I think this has bitten me before, thanks!

PR #52, opened 2020-09-15
add .desktop file

KottV:
This file adds BespokeSynth shortcut in Linux Desktop Environments.

### DIFF
--- a/BespokeSynth.desktop
+++ b/BespokeSynth.desktop
@@ -1,0 +1,10 @@
+[Desktop Entry]
+Name=BespokeSynth
+GenericName=Software modular synth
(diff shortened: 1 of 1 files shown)


awwbees:
thanks! when I distribute linux builds, where should this file go? next to the executable?

KottV:
it must be placed in
/usr/share/application/
and icon in the
/usr/share/icons/hicolor/512x512/apps/

KottV:
or, if this is not system-wide package, you can place it in user's home:
~/.local/share/application/ and ~/.local/share/icons/hicolor/512x512/apps/

(of course path to binary must be changed)

PR #59, opened 2020-10-29
Added check for isSysEx for OnMidi

bgaster:
SysEx messages don't contain channel information and so cannot be filtered on a per channel basis. To enable sending thru sysex messages, even when channel filtering is enabled, I added a check to allow sysex messages to be forwarded on.


awwbees:
excellent, thanks!

PR #62, opened 2020-11-12
change c++ to 14, add -fvisibility=hidden

KottV:
Hi,

It fails to compile with recent changes:
```
[  791s] ../../Source/BiquadFilter.cpp: In member function 'float BiquadFilter::GetMagnitudeResponseAt(float)':
[  791s] ../../Source/BiquadFilter.cpp:195:21: error: use of 'auto' in lambda parameter declaration only available with '-std=c++14' or '-std=gnu++14'
[  791s]   195 |    auto square = [](auto z) { return z * z; };
```

also i've added -fvisibility=hidden which reduces amount exposed symbols on Linux (https://github.com/kushview/Element/issues/197)


awwbees:
ah thank you! strange that MSVC was fine with me using c++14 stuff even though I accidentally still had it set to c++11, but I guess that's compilers for ya.

Issue #12, opened 2020-03-13
system-wide install and data path


KottV:
Hi,

I'd like to package BespokeSynth (I hope, you're not against more popularity) in Linux repositories.
One thing stops that - relative path for 'data' dir to binary. BS won't start without minimal configuration, so it has to be somewhere in system, if we gonna to install it via package manager.
What do you think, is it possible to distinguish factory and user datadir?
JUCE has File::commonApplicationDataDirectory File::userApplicationDataDirectory File::globalApplicationsDirectory File::userHomeDirectory functions which helps to find userdir and system path for data storing.

From my side I'll try to realize using examples from other JUCE synth (with my zero-level skills in c++)

awwbees:
sorry, I missed this message! I don't know much about packaging in linux repos, but I'm definitely not against it. I'll have a look to figure out a way to have a system-wide install path... maybe I'll just have a priority system where bespoke scans through a list of directories when it starts up, and stops at the first valid one it finds

KottV:
Ok, it's not so important feature for now)

KottV:
As linux users are here now and I'm on way do make .deb packages in OBS, I'd like to rise up this question again. For now I package Bespoke with these modifications:
```
#elif JUCE_LINUX
   static string sDataDir = "";
   if (sDataDir == "")
   {
//      string localDataDir = File::getCurrentWorkingDirectory().getChildFile("data").getFullPathName().toStdString();
      string localDataDir = File::getSpecialLocation(File::userApplicationDataDirectory).getChildFile("BespokeSynth/data").getFullPathName().toStdString();
      if (juce::File(localDataDir).exists())
         sDataDir = localDataDir;
      else
//         sDataDir = File::getCurrentWorkingDirectory().getChildFile("../../MacOSX/build/Release/data").getFullPathName().toStdString();   //fall back to looking at OSX dir in dev environment
         sDataDir = File::getSpecialLocation(File::globalApplicationsDirectory).getChildFile("share/BespokeSynth/data").getFullPathName().toStdString();
   }

   return sDataDir + "/" + path;
#else
```

It makes to look data in ~/.config/BespokeSynth and fallback to /usr/share/BespokeSynth
@djdeath @awwbees what do you think about that concept?

awwbees:
that seems like a great concept! I probably would want to modify it to:
```
#elif JUCE_LINUX
   static string sDataDir = "";
   if (sDataDir == "")
   {
[...]

PR #91, opened 2021-05-24
add linux default prefs, enable Jack, disable curl

KottV:
Hi,

This adds userprefs file for Linux, also enables Jack audio support and removes curl dependency


KottV:
I wrote first attempt to address https://github.com/awwbees/BespokeSynth/issues/12

Source/UserData.cpp
```diff
@@ -0,0 +1,53 @@
...
+    #endif
+    juce::File localConfig (File::getSpecialLocation(File::userApplicationDataDirectory).getChildFile("BespokeSynth/data/userprefs.json").getFullPathName());
+    juce::File localBlank (File::getSpecialLocation(File::userApplicationDataDirectory).getChildFile("BespokeSynth/data/layouts/blank.json").getFullPathName());
+    
+    juce::File localConfigSave (File::getSpecialLocation(File::userApplicationDataDirectory).getChildFile("BespokeSynth/data/userprefs_save.json").getFullPathName());
+    juce::File localBlankSave (File::getSpecialLocation(File::userApplicationDataDirectory).getChildFile("BespokeSynth/data/layouts/blank_save.json").getFullPathName());
+    
+    if(juce::File(localDataDir).exists())
+    {
+        localConfig.copyFileTo(localConfigSave);
```
KottV, line 22: I assume that `userprefs.json` and `layouts/blank.json` are user-modified files.

KottV:
Nah, it doesn't work on JUCE 5.4.7

awwbees:
I am taking a look at this. apologies for the slowness! I am pretty inexperienced with git/github and am still figuring out how to test and merge pull requests properly

KottV:
I see!

You can make a `test` or `develop` branch where we can test commits.
Anyway this addition is completely safe :)

KottV:
honestly this is my first JUCE code and of course it's better if you'll take a steady look on it

awwbees:
ok! I've got a few changes I'm going to make, to better support having this behavior cross-platform like you had suggested in #12 

one question about a change in here: what was the purpose of disabling curl? does it cause some specific complication in linux?

KottV:
>what was the purpose of disabling curl? does it cause some specific complication in linux?

dependencies in linux are headache
in Ubuntu it's libcurl-gnutls, in other distributives - libcurl, binary compiled in Ubuntu wont run on SUSE or Fedora without tricks
as curl is not really needed it is worth to just disable it

PR #97, opened 2021-07-22
fix Linux linking on Ubuntu 21.04

djdeath:
For some reason the Debug build is not linking on my machine (I haven't tried release build yet).

First python-config is gone, it's either python2-config or python3-config

There are missing python symbols, that's an easy fix with --embed to python3-config.

And finally the linker is complaining about position independent stuff, adding -fPIE fixes the issue.


awwbees:
thanks!

KottV:
`python3-config` it won't work for Debian 11 and even Ubuntu 20

PR #100, opened 2021-07-27
M185Sequencer: tweak behavior when disabled

djdeath:
Currently when disables, the module stops ticking and doesn't stop the
last note played.

This change keeps it on tick and stops the last note.


djdeath:
Oops, sorry, I just noticed you renamed Open to Repeat.

Give me a day and I'll update this to use your changes.

djdeath:
Should be good now.

PR #102, opened 2021-08-23
Fix python3 Linux linking flags

KottV:
Hi,

It works for openSUSE, Ubuntu 20.04 and Debian 11 (python3.9)


KottV:
Ohhh, it seems it works only for python >= 3.8 and doesn't for 3.6 :(

awwbees:
ah, that is fine, I think python 3.8 will be my minimum supported version anyway. that mirrors the requirements on windows and mac. I inadvertently only have 3.6 installed on my linux machine, I need to update.

awwbees:
I'm working through some issues getting my machine configured correctly for python3.8, I will merge this PR once I resolve that and can test it

KottV:
no problem, i've a patch in my build repo, with condition for 3.6 and 3.8 python versions

awwbees:
ah... should I merge this change, or do you have another change that works better? I just updated to ubuntu 20.04 (which was the best way I could see to get python 3.8) and tested this change, it worked for me.

KottV:
yes, it's a proper solution (i'm sure as hell!)

KottV:
Also added JACK Client name

awwbees:
thanks!

PR #103, opened 2021-08-25
user another trick to pass JACK Client name, suitable for 5.4.7

KottV:
previous worked in Projucer 6.0.8 only
quoting seems tricky in 5.4.7
i think adding that in  AppConfig.h user's sections is the "best" workaround for now


awwbees:
aha, thank you for sorting this out

PR #104, opened 2021-08-31
fix build with gcc-11

KottV:
Hi,

It fixes compiling with gcc-11 and recent updates


awwbees:
I will take a look at this once I am back home so I can look on all platforms. Thanks for the fix!

PR #140, opened 2021-09-16
Improve sample player user experience

cassaundra:
This PR makes three changes:
- When hovering over cue points in the sample player, select the cue point closest to _but not exceeding_ the cursor position. This fixes an issue where a cue point cannot be selected if it overlaps with another cue point of a lower index.
- Use exponential scroll zooming. I'm looking for feedback on this, but it feels more natural to me and in line with other audio editors.
- Fix sample player lag caused by repeated system clipboard checks.

Thank you!

### DIFF
--- a/Source/CodeEntry.cpp
+++ b/Source/CodeEntry.cpp
@@ -25,6 +25,7 @@
   ==============================================================================
 */
 
+#include "ModularSynth.h"
 #if BESPOKE_WINDOWS
 #define ssize_t ssize_t_undef_hack  //fixes conflict with ssize_t typedefs between python and juce
 #endif
@@ -933,7 +934,7 @@ void CodeEntry::OnKeyPressed(int key, bool isRepeat)
       if (mCaretPosition != mCaretPosition2)
          RemoveSelectedText();
       
-      juce::String clipboard = SystemClipboard::getTextFromClipboard();
+      juce::String clipboard = TheSynth->GetTextFromClipboard();
       AddString(clipboard.toStdString());
    }
    else if (toupper(key) == 'V' && (GetKeyModifiers() == (kModifier_Command | kModifier_Shift)))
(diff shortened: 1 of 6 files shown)


cassaundra:
I'm noticing some fairly significant lag caused by the polling clipboard check on Linux/X11 when multiple sample players are active. I'm curious to know if anybody else can reproduce this.

If this is the case, the most unobtrusive fix would be to only check the clipboard for a single sample player on mouse enter. I'll go ahead and implement this, awaiting feedback. :wink:

https://github.com/awwbees/BespokeSynth/blob/94a841bee6882a6e4b16bf627bf26cafece6d096/Source/SamplePlayer.cpp#L174-L198

cassaundra:
Commit 051d28bf46d4fdbd1d8526299f7df4ba67043253 fixes the issue of lag caused by multiple sample players trying to access the system clipboard at the same time, by shifting that responsibility to the `ModularSynth` class (see commit details for more information). The behavior is essentially unchanged, only without the lag.

awwbees:
thank you for this work!! sorry that it took me several days to look at it

- the cue highlighting is much better than what I had.
- I have not experienced the clipboard issue myself (I don't use linux much), but I have gotten reports of it from other people, so it is nice to fix that
- my only issue is with the zoom: I thought the old zoom felt better, this version feels fast and inaccurate to me. could you remove that? but if you feel strongly about it, maybe open it as a separate pull request, and solicit feedback from additional users

awwbees:
oh also: could you rename ModularSynth::clipboard to ModularSynth::mClipboard, to follow the naming conventions?

cassaundra:
Thank you for your reply! I went ahead and renamed `ModularSynth::clipboard` to fit the naming conventions (oops), and reverted the change to the sample player's zoom. The PR should be ready to be merged now, but let me know if there are any other changes to be made.

awwbees:
excellent, thank you for the contribution!

PR #147, opened 2021-09-17
Improve Linux build scripts

4oo4:
I set this to bash for now, but it can always be changed to sh later. I wasn't sure about the MacOS shell scripts, but you could do the same thing there (if `env` works the same way?) or change these to `#!/bin/sh` to match, even though I think using `env` is best practice for Linux.


dromer:
If you're going to fix these files then maybe also make sure there are newlines at the end of them.

(although honestly the whole build system could use a revamp to be more modern and robust. TBD)

4oo4:
Done, good catch. Agreed that the build system could be a lot better (I think what it really needs is CI), but this should at least make it a tiny bit better for now.

baconpaul:
Hi linxuers! Comments and help on #187 appreciated if you want.

dromer:
Wow, awesome work @baconpaul !

I'll try your build right away!

4oo4:
Awesome, that's a way better solution so I'm closing this.

PR #174, opened 2021-09-18
Fix #142: time signature quantization with kInterval_CustomDivisor

### LINKED ISSUE #142: Circle sequencer uneven depending on time signature
The circle sequencer behaves oddly on time signatures where the top and bottom number don't match. It seems like the subdivisions of a ring always act as if it were 4/4. I'd expect once around the ring to always represent a measure. I noticed this initially in 9/8 but got weirdness in every time signature I tried except 2/2, 4/4, 8/8, and 16/16.

Steps to reproduce a simple case:

* Set time signature to 3/4
* Add a circle sequencer and set length for a ring to 3
* Add notes on that ring
* Those notes don't trigger evenly

Example image:
image
The yellow ring has 4 divisions but is playing 3 evenly spaced notes on the stream. The blue ring never plays I assume because it's the 4th quarter note in a measure of 3/4. Similarly the top row/purple ring should be 3 evenly spaced quarter notes but instead looks like a 4/4 triplet stuffed into a bar of 3/4.

Here's how the note stream for that example looks in 4/4. I'd expect it to look mostly like this no matter what the time signature is.
image

Recording with that pattern on a few different time signatures: https://streamable.com/nwt3ww

iamevn:
Fixes #142
Example from issue with this patch: https://streamable.com/7h21zr

### DIFF
--- a/Source/Transport.cpp
+++ b/Source/Transport.cpp
@@ -339,7 +339,6 @@ int Transport::GetQuantized(double time, const TransportListenerInfo* listenerIn
    int measure = GetMeasure(time);
    double measurePos = GetMeasurePos(time);
    double pos = Swing(measurePos);
-   pos *= double(mTimeSigTop) / mTimeSigBottom;
    
    NoteInterval interval = listenerInfo->mInterval;
    switch (interval)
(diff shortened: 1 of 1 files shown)


awwbees:
excellent!! thanks. I obviously never tested custom divisors in odd time!

PR #186, opened 2021-09-18
Disable non-main output busses for now

### LINKED ISSUE #148: VST3 Busses and why Surge doesn't work in Bespoke
Hi!

Surge XT (our latest version) doesn't work in Bespoke. It loads and scans but says the audio is never called for, because you actually don't ever call ::processBlock 

The reason for this, though, is because of our bus layout. Surge is a synth (midi in) but also supports one audio in and up to 3 audio outputs. It works fine with no audio in and one stereo out of course but the other configurations are valid

If I rebuild surge to not have this (namely, take the extra output busses out of the picture) then bespoke runs surge xt just fin, and is fun and cool. 

But somewhere you are convincing yourself i'm not valid because of my 6 output configuration. Can't quite see it in the source.

The juce api getBusLayout and so on is implemented properly in surge. Hit my up here or on discord if you need any questions. I can't quite get the thing to build on mac and your exe won't let me attach a debugger alas, so i'm a bit stuck to figure out more than this, but if you build surge at head and comment out the scene a "withOutput" in SurgeSynthProcessor, it all works fine. SO that's definitely the problem.

Thanks!

baconpaul:
Until Bespoke gets multi-output support, synths which have it
(SurgeXT, Pianotec, some samplers) need the aux busses disabled
and just to have the main. This code accomplishes that.

Closes #148

### DIFF
--- a/Source/VSTPlugin.cpp
+++ b/Source/VSTPlugin.cpp
@@ -328,6 +328,15 @@ void VSTPlugin::LoadVST(juce::PluginDescription desc)
    if (mPlugin != nullptr)
    {
       mPlugin->enableAllBuses();
+
+      // DIsable all non-main output busses
+      auto layouts = mPlugin->getBusesLayout();
+
+      for (int busIndex = 1; busIndex < layouts.outputBuses.size(); ++busIndex)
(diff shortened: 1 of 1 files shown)


awwbees:
thank you!

PR #205, opened 2021-09-20
Mac Resource Location Changes

### LINKED ISSUE #117: Mac version is tied to a hard-coded, non-standard path
The Mac version of the application REQUIRES that the resources directory be located in /Applications/BespokeSynth/resources.  Not only does this both require and lock down the BespokeSynth folder location in a place that should never require any folders at all, but it also badly violates Mac file system standards in two ways.  First (1) the Applications directory is supposed to hold *applications* (that is, .app), and (2) applications and their respective folders should be able to be moved to anywhere the user sees fit, including (of course) ~/Applications/

I suggest one of two fixes.  Either:

- Tuck the resources directory inside the BespokeSynth.app directory and refer to it using *relative* path references in the application proper, or

- Install the resources directory in /Library/ApplicationSupport/BespokeSynth/

The first is strongly preferable.

baconpaul:
1. The 'resource' directory lives in the bundle and CMake copies
   them at build time
2. The cmake commands can create a DMG
3. That DMG contains just the app, which runs from the DMG
   with no install

Closes #117

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -438,6 +438,26 @@ if (APPLE)
             )
     target_link_libraries(BespokeSynth PRIVATE
             "-framework CoreAudioKit")
+
+    # Post install tasks - copy resources to the bundle
+    get_target_property(OUTDIR BespokeSynth RUNTIME_OUTPUT_DIRECTORY)
+    add_custom_command(TARGET BespokeSynth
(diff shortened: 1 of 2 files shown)


baconpaul:
Closing this - will reopen with a modified branch shortly.

PR #213, opened 2021-09-20
Resource Cmake Changes; Mac Resource Location Changes

### LINKED ISSUE #117: Mac version is tied to a hard-coded, non-standard path

baconpaul:
On Win/Lin:
- To make dev easier, the resource directory is copied next
  to the dev exe on win and lin, and the existing local
  relative resource directory scan works, avoiding all the
  ../../../ stuff being hardcoded to relative paths

On Mac:
- The 'resource' directory lives in the bundle and CMake copies
  them at build time
- Use CFBundle API to find our bindle and resolve the path
- This defacto means the mac app is fully senf contained and
  can be copied anywhere.

Closes #117

cml

rb

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -474,6 +474,9 @@ elseif (WIN32)
     endif ()
 
     # Post install tasks - copy resources to the exe output directory
+    set_target_properties(BespokeSynth PROPERTIES VS_DEBUGGER_WORKING_DIRECTORY "$<TARGET_FILE_DIR:BespokeSynth>"
+                                    VS_DEBUGGER_COMMAND           "$<TARGET_FILE:BespokeSynth>"
+                                    VS_DEBUGGER_ENVIRONMENT       "PATH=%PATH%;${CMAKE_PREFIX_PATH}/bin")
     get_target_property(OUTDIR BespokeSynth RUNTIME_OUTPUT_DIRECTORY)
     add_custom_command(TARGET BespokeSynth
             POST_BUILD
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ if (NOT DEFINED PYBIND11_FINDPYTHON)
     endif()  
 endif()
 add_subdirectory(libs/pybind11 CONFIG)
+message(STATUS "Bespoke: Python Executable is ${PYTHON_EXECUTABLE}")
 
 if (NOT BESPOKE_JUCE_LOCATION)
     set(BESPOKE_JUCE_LOCATION "libs/JUCE" CACHE STRING "" FORCE)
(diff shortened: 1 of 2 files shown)


awwbees:
hmm, I'm gonna look at this further, but I'm not so sure about having two copies of the resources dir: it seems like it would make it too easy to accidentally modify the wrong copy of a file while doing development that involves changing the contents of the resources dir.

is there a better way to figure out the relative path of the resources dir from the executables that cmake/visual studio/xcode spit out? maybe just ensure that all of various build methods put the executable at the same relative depth, even if that means introducing unnecessary layers of directories? I will take a look.

awwbees:
ahh, ok, I think I misunderstood about when the copy happens. this is all good on mac, gonna test windows next!

baconpaul:
Yeah the copy happens after every build so the repo resources is the master 

if you edit it in place you need to remember to copy back to repo

basically the strategy is ./resources is master and gets copied to bundle for distribution on Mac and to build directory for convenience on win and lin

awwbees:
okay, further testing:
- mac works both when building via cmake and xcode
- windows works when building via cmake, but fails to find the resources directory when launching from visual studio

the visual studio issue seems to be that the working directory is set to $(ProjectDir) instead of $(TargetDir). when I manually set it to $(TargetDir), it works. quick googling seems to indicate that there is not a simple cmake way to automatically fix this (though maybe I'm looking at out of date information) but there are some workarounds. looking more now.

image

awwbees:
I think this snippet fixes it, testing now:
```
set_target_properties(BespokeSynth PROPERTIES VS_DEBUGGER_WORKING_DIRECTORY "$<TARGET_FILE_DIR:BespokeSynth>"
                                                  VS_DEBUGGER_COMMAND           "$<TARGET_FILE:BespokeSynth>"
                                                  VS_DEBUGGER_ENVIRONMENT       "PATH=%PATH%;${CMAKE_PREFIX_PATH}/bin")```

awwbees:
ok, yeah, that snippet worked!

baconpaul:
Oh that’s awesome! I will add that to this pr in a little bit (and as mentioned offline going forward send my prs so you have edit access)

awwbees:
awesome, thank you!

baconpaul:
Yay! Great

OK I'll find some time this week to make the linux changes we discussed on discord and then i think we will be in pretty good shape.

PR #215, opened 2021-09-20
cable splitting documentation improved

severak:


awwbees:
this text ends up overflowing offscreen in the in-app help display, and I think it's a bit verbose for the basic help. I'll try to think of a different way to communicate this.

PR #216, opened 2021-09-20
Open state files on app startup

### LINKED ISSUE #209: Associate .bsk files with bespoke on mac
If I downloaded a patch, I'd like to be able to open it and have bespoke load it.
Currently:
image
Even if trying to choose application and selecting bespoke, it isn't reccomended:
image
And even after switching from "Reccomended applications" to "All applications", trying to open the file with bespoke opens the default starting patch.  
I don't know the status with other operating systems, but this is the state on mac.

Tadaboody:
This pull request allows to 

- Open state files on app startup
- Open state files by dragging into an existing 

Opening as a draft because there seems to be a bug with the "Reset Layout" button not doing anything when opening a file from startup, and because I've only tested on MacOS

Closes #209

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -390,6 +390,7 @@ juce_add_gui_app(BespokeSynth
         BLUETOOTH_PERMISSION_ENABLED    TRUE
         FILE_SHARING_ENABLED            TRUE
         DOCUMENT_BROWSER_ENABLED        TRUE
+        DOCUMENT_EXTENSIONS             .bsk
         )
 
 juce_generate_juce_header(BespokeSynth)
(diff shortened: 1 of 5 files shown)


jpcima:
Not a developer of this, but interested in feature offered by this PR.
However it has lots of issues.

1. file is not opened from command line. `anotherInstanceStarted` never called, but it's fine inside `initialise`
2. endsWith: `std::string_view` is missing, not building in c++ 17 mode
    suggestion: remove and use `juce::String(files[0]).endsWith()`
3. use `getCommandLineParameterArray` instead of `commandLine`, it prevents errors with use of spaces
   See also https://forum.juce.com/t/prob-user-error-juceapplication-initialise-breaks-arguments-with-spaces/26686

Also the indentation style looks inconsistent with the other code.

awwbees:
thank you for this! I took these changes and got them working in #424 

(there might have been a way for me to directly get it to work via this PR, but my git-fu is too weak...)

PR #217, opened 2021-09-21
Flip offset values for circle sequencer

### LINKED ISSUE #121: Circle sequencer offsets are inverted
Increasing the offset value for a ring in the circle sequencer moves the offset backwards in time, and decreasing it moves it forward. This only affects the note itself; the UI correctly moves the rings forwards or backwards, causing the UI to not reflect the notes as actually played.

Issue encountered using the Windows build of version 1.0.0 on Windows 10.

iamevn:
flip visuals, actual timing seemed consistent with other modules

fixes #121 

recording of offset behavior with this patch applied (compares with the offset on the random note module): 
https://streamable.com/3e5ikb

### DIFF
--- a/Source/CircleSequencer.cpp
+++ b/Source/CircleSequencer.cpp
@@ -234,7 +234,7 @@ void CircleSequencerRing::Draw()
    ofFill();
    for (int i=0; i<mLength; ++i)
    {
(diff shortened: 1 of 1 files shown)


awwbees:
nice, simple change. thanks!

PR #229, opened 2021-09-22
Fix subpixel rendering issue in quick spawn menu

cassaundra:
When there are an odd number of items in the quick spawn menu, the height of the module ends up having a fractional half left over, resulting in a subpixel offset that makes the text render blurry on some displays. This PR fixes that issue by truncating the fractional part of the height and width of the quick spawn menu when it is first shown. The effect is shown below:

<p align="center">
  
  
</p>

### DIFF
--- a/Source/QuickSpawnMenu.cpp
+++ b/Source/QuickSpawnMenu.cpp
@@ -29,6 +29,7 @@
 #include "ModularSynth.h"
 #include "ModuleFactory.h"
 #include "TitleBar.h"
+#include <cmath>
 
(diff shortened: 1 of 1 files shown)


cassaundra:
This issue still remains for non-integer `ui_scale` config values in the quick spawn menu and other UI elements, which leads me to think that there's a better way to solve this issue at a more fundamental level. If that's the case, I'll just go ahead and close this PR.

awwbees:
I think there's a bigger subpixel font issue that I have to work out... I will hold off on this one for a more systemic fix

PR #236, opened 2021-09-22
Make Linux Bespoke aware of CMAKE_INSTALL_PREFIX

baconpaul:
This commit accomplishes three things

1. Sets up a rudimentary Cmake-to-CPP constant communication by having
   a VersionInfo file (but don't do the full surge thing of git versions
   and the like yet)
2. With that done, modify the linux resource location waterfall to be
   a. working directory / resoruce
   b. $ENV{BESPOKE_DATA_DIR}
   c. ${CMAKE_INSTALL_PREFIX}/resource
   d. /usr/share/BespokeSynth/resource
3. With that done, create a cmake install rule so a succesfull build
   and install on linux is now

```
cmake -Bignore/lin -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/what/ever/you/want
cmake --build ignore/lin --config Release
sudo cmake --install ignore/lin
```

and after that then `/what/ever/you/want/bin/BespokeSynth` will be installed and
load resources from `/what/ever/you/want/share/BespokeSynth/resources`

Tested on ubuntu 20.

### DIFF
--- a/Source/VersionInfo.h
+++ b/Source/VersionInfo.h
@@ -1,0 +1,7 @@
+#include <string>
+
+namespace Bespoke
+{
+    extern const char* VERSION; // This will be the same string as Juce::Application::getApplicationVersion()
+    extern const char* CMAKE_INSTALL_PREFIX;
+}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ if (NOT CMAKE_BUILD_TYPE)
 endif ()
 
 
-project(BespokeSynth VERSION 0.9.0 LANGUAGES C CXX ASM)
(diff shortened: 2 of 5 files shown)


awwbees:
cmake is so cool! thanks.

Source/VersionInfo.h
```diff
@@ -0,0 +1,7 @@
+#include <string>
+
+namespace Bespoke
+{
+    extern const char* VERSION; // This will be the same string as Juce::Application::getApplicationVersion()
+    extern const char* CMAKE_INSTALL_PREFIX;
+}
```
dromer, line 7: oops, missing newline :#

Issue #151, opened 2021-09-17
Highlights don't always match click targets


ianh:
If you hover your mouse just outside the top left enable/disable toggle, the blue highlight ring appears.  But clicking here doesn't actually enable or disable the module.  There may be other places this happens too.

ianh:
Yeah, this happens nearly everywhere there's a highlight ring.  I think fixing this would make the UI feel way better!

ianh:
Nice, looking forward to trying this out!

PR #240, opened 2021-09-22
Reset the hovered UI control immediately when the cursor leaves its bounds

ianh:
When a control (particularly a button) is highlighted, it creates the expectation that clicking will interact with that control.  Leaving the highlight active within a 10 pixel boundary does make things less visually jittery, but it defeats this fundamental UI expectation.

Fixes issue #151.


ianh:
I have a better patch for this -- new PR incoming.

PR #241, opened 2021-09-22
Use a timeout instead of a margin for hover hysteresis

ianh:
There are two parts to this patch:

1. When a control (particularly a button) is highlighted, it creates the expectation that clicking will interact with that control.  Leaving the highlight active within a 10 pixel boundary does make things less visually jittery, but it defeats this fundamental UI expectation.  So this patch disables the highlight immediately when the mouse leaves the control's bounding box.

2. The visual highlight isn't the only effect of the hover state.  There's also a tooltip and some descriptive text that appears in the top left.  These effects don't need to be as precise as the visual highlight, and in fact, making them overly precise leads to distracting flashing in and out as you move your mouse over a dense arrangement of controls.  To avoid this flashing, keep the hover state active for a few frames after the mouse leaves the control.  Using a timeout rather than a margin avoids the confusing situation where you're hovering the mouse a few pixels outside of the control, and it's not highlighted, but the tooltip is staying around anyway.

Fixes issue #151.

### DIFF
--- a/Source/IUIControl.cpp
+++ b/Source/IUIControl.cpp
@@ -86,6 +86,8 @@ void IUIControl::CheckHover(int x, int y)
    if (TheSynth->GetFrameCount() != sLastHoveredUIControlFrame && TestHover(x, y) && (gHoveredUIControl == nullptr || !gHoveredUIControl->IsMouseDown()))
    {
       gHoveredUIControl = this;
+      gHoveredUIControlIsHighlighted = true;
+      gHoveredUIControlFramesUntilExpiration = 0;
       sLastHoveredUIControl = this;
       sLastUIHoverWasSetViaTab = false;
       sLastHoveredUIControlFrame = TheSynth->GetFrameCount();
@@ -97,7 +99,7 @@ void IUIControl::DrawHover(float x, float y, float w, float h)
    if (Push2Control::sDrawingPush2Display)
       return;
    
-   if (gHoveredUIControl == this && IKeyboardFocusListener::GetActiveKeyboardFocus() == nullptr && TheSynth->GetGroupSelectedModules().empty())
+   if (gHoveredUIControl == this && gHoveredUIControlIsHighlighted && IKeyboardFocusListener::GetActiveKeyboardFocus() == nullptr && TheSynth->GetGroupSelectedModules().empty())
    {
       ofPushStyle();
       ofNoFill();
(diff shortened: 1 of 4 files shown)


awwbees:
this is really good feedback about the hovering and expectation of clicking, and how it affects the feel of the UI. I think a little more thought is needed though...

the hover interaction actually has a lot of usages where the current "sloppy hover" is desired. here are a few:
- while a control is hovered, you can scroll the mouse wheel to adjust the value
- while a control is hovered, typing allows you to manual enter a value
- while a control is hovered, moving a midi controller allows you to bind the hovered control to that knob

all of these interactions benefit from the current "sloppy hover", and I think could feel too finicky if the system was changed from a spatial hysteresis to a timing hysteresis. suffice it to say, the intent of the spatial hysteresis system is not to reduce visual jitter, but instead to achieve these goals.

probably unintuitively, the hover state actually has no connection to click.

I wonder if an inverse solution to #151 would be more appropriate? if clicking when a control is hovered always counts as a click on that control (with the click position clamped to the control's dimensions). I will have to think about this one. thank you for your passion in wanting to clean up bespoke's rough edges.

ianh:
Ah, thanks for the quick response and detailed explanation!  I hadn't realized the larger context of this behavior.  The inverse solution seems like it could work.  Adding this code to `void ModularSynth::MousePressed(int intX, int intY, int button)`:

```c++
   if (gHoveredUIControl != nullptr)
   {
      IDrawableModule *parent = gHoveredUIControl->GetModuleParent();
      float x = GetMouseX(parent->GetOwningContainer());
      float y = GetMouseY(parent->GetOwningContainer());
      float uiX, uiY;
      gHoveredUIControl->GetPosition(uiX, uiY);
      float w, h;
      gHoveredUIControl->GetDimensions(w, h);
      if (x < uiX)
         x = uiX;
      if (y < uiY)
         y = uiY;
      if (x > uiX + w)
         x = uiX + w;
      if (y > uiY + h)
         y = uiY + h;
      if (parent->TestClick(x, y, rightButton))
         return;
   }
```

...and playing around with it a bit, the behavior feels good to me, except in a few cases where there are controls near interactable elements that don't have a hover state (like the lookahead/autosave/... controls near the instruments/note effects/synths/... drop downs when the window is narrow) -- hovering the control and moving the cursor over the element without a hover state, the control remains hovered and will continue accepting clicks.  Anyway, I'll close this PR and let you decide what you'd like to do here.  Thanks for making Bespoke, it's a lot of fun!

awwbees:
thank you! yeah, the code you posted is exactly what I had in mind. I'll try it, and then see if I can figure out how to iron out the issues you mention (plus all of the other issues that are bound to spring up)

PR #248, opened 2021-09-23
Fix segfault when writing data to file

tomc1998:
There are 2 fixes here:

1. a `char*` (`path`) was being passed into a function expecting a `std::string` (unsure why this wasn't a compiler error, who knows)
2. The raw pointer inside `outputTo` was being passed into `wavFormat->createWriterFor`. However, this actually takes ownership of the pointer & frees it at the end of the scope, leading to a double free when `outputTo` tries to free itself. I believe `.release()` stops the `unique_ptr` freeing itself, I'm no C++ guru though :p


baconpaul:
@awwbees FYI: the char* is a bit of a red herring but the second fix is dead on correct. `createWriterFor` takes ownership of the stream and when I ported to 6.1.1 i mis-read the API so there's a double free here.

Thanks @tomc1998 !!!

awwbees:
thank you @tomc1998!!

PR #249, opened 2021-09-23
A rudimentary azure pipelines script

baconpaul:
This is a rudimentary azure pipelines script which will do PR
builds and main merge builds but not put them anywhere yet and
always be debug and the like. But a starting point!

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,102 @@
+# BespokeSynth Azure Pipelines script
+
+trigger:
+  - main
+
+pr:
+  - main
+
+jobs:
+  - job: Build
+    variables:
+    strategy:
+      matrix:
+        macOS-x64:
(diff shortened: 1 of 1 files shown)


baconpaul:
OK @awwbees to get things going with azure

1. Merge this. It may or may not be right but without an azure pipeline file you can't even invite azure
2. if you don't have an msft account, create one and then
3. follow https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml to invite azure to your repo

that should in theory give us PR builds. But if it splats I can do another PR to test it once azure is running :)

Once that works we can have a separate step to get nightly binaries up and going. This pipelines file won't do that but i have one which will of course :)

awwbees:
sweet, completely amazing!!! I will get this set up this evening.

awwbees:
here goes nothing!

PR #256, opened 2021-09-24
 translated tooltips into spanish , needs a revision 

CarloCattano:
Could be nice to have it revised by several native speakers from different Spanish speaking countries , but its quite tedious to just read it so my guess is we can test it as it goes and then its there for anybody to modify via pull request

### DIFF
--- a/resource/tooltips_esp.txt
+++ b/resource/tooltips_esp.txt
@@ -1,0 +1,1797 @@
+
+
+
+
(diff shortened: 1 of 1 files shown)


awwbees:
oh wow. this is incredible. I can't believe that all the tooltips are localized to spanish only ten days after release. thank you so much!

Issue #246, opened 2021-09-23
"CTRL+A" to select all inside text


ZachIsAGardner:
When I'm editing the text inside a module and I input "CTRL+A" I'm expecting it to highlight all the text. Instead it inputs an "A" like I'm holding down shift.

Can we make it so when you do "CTRL+A" it highlights all the text? This would be pretty useful for when renaming things or typing out a comment since it starts with "insert comment here".

fredericlb:
Hi ! I started working on my first PR to add selection in text inputs. If no one is working on this it would be pretty easy for me to add main keyboard shortcuts like this one.

awwbees:
that would be awesome @fredericlb! I have code that does that in CodeEntry.cpp, and never got around to bringing it into TextEntry. if you haven't looked at the CodeEntry stuff yet, it could be a useful resource!

fredericlb:
Thanks, I started writing my own stuff to handle selection but after your comment the PR is now mainly a copy/paste of what was in CodeEntry :)

awwbees:
got added in the PR referenced above

PR #258, opened 2021-09-24
Add selection features to TextEntry + Small CodeEntry fix

fredericlb:
Mostly a copy of CodeEntry selection related code to TextEntry, including main shortcuts support. 
This PR also fixes the caret position when it is moved out of selection (it should be at the beginning or end of the selection depending on which key was pressed). 

Solves #246 .

This is my first PR here and it's been a long time since I haven't worked on a C++ project, so tell me if anything is wrong/missing in my submission and I will fix it.

### DIFF
--- a/Source/CodeEntry.cpp
+++ b/Source/CodeEntry.cpp
@@ -849,7 +849,9 @@ void CodeEntry::OnKeyPressed(int key, bool isRepeat)
       }
       else
       {
-         if (mCaretPosition > 0)
+         if (!(GetKeyModifiers() & kModifier_Shift) && mCaretPosition != mCaretPosition2) 
+            MoveCaret(MIN(mCaretPosition, mCaretPosition2));
+         else if (mCaretPosition > 0)
             MoveCaret(mCaretPosition - 1);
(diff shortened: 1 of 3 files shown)


awwbees:
awesome, thank you, I will take a look at this soon!

(please ignore the "failed checks", the azure stuff is a work in progress)

awwbees:
I hit an issue that I haven't investigated yet, but here is a repro:

1. click into a text entry (such as the one that says "440" on the scale module)
2. use shift + arrow keys to select the "40" part
3. type the number 1. at this point, I expect the text entry to read "41", but instead it reads "4140"
4. further issue: from this bugged state, hit backspace repeatedly. bespoke will crash, presumably because some expected state is out of sync

I will be able to look at this in more depth later, but if you get to it first, that would be great!

fredericlb:
I will investigate this one tomorrow, I think I've guessed what's happening :)

fredericlb:
this commit should fix the use case (and the backspace related crash too)

awwbees:
thank you!!

PR #268, opened 2021-09-25
More Mac Python Stuff

baconpaul:
Striping that python signing was great, but it means the default
signing at the end of xcode breaks, so

1. After i strip, resign with identity '-' (and we can clean
  this up at packaging time later) and
2. Make the pipeline build on mac use Xcode for now


awwbees:
it worked! I built via xcode and ran successfully

PR #278, opened 2021-09-27
Expanded Microtonal Support

baconpaul:
1. Fix a couple of calculation errors with the built in
   scales
   - the Pythag had an incorrect value at F#
   - the retuned scales ignored the fixed pitch at note setting
   - This results in pythag with note 69 at 440 tuning ntoe 69 to 440
     and also restores the more classic 'wolf howl' at F# fifth

2. Add rudimentary support for ODDSound MTS (https://oddsound.com)
   - If you put yourself in oddsound scale mode it will query an
     MTS Master if you have ome running
   - Mac and Win only on this feature currently even though it
     compiles on LIN
   - binds to the oddsound client library implementation in libs/

3. Add rudimentary support for direct SCL/KBM file load
   - Scale object can load either file and adjust tuning aacordingly
   - Scale object streams the file contents into the bsk and restores
     appropriately
   - Binds to the surge tuning-library implementation in libs/

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,9 @@
 [submodule "libs/pybind11"]
 	path = libs/pybind11
 	url = https://github.com/pybind/pybind11
+[submodule "libs/tuning-library"]
+	path = libs/tuning-library
+	url = https://github.com/surge-synthesizer/tuning-library
+[submodule "libs/oddsound-mts/MTS-ESP"]
+	path = libs/oddsound-mts/MTS-ESP
+	url = https://github.com/ODDSound/MTS-ESP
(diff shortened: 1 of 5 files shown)


baconpaul:
This adds new submodules; if you are running and pull the branch down you willl need to do a `git submodule update --init --recursive` to grab them.

awwbees:
could you update this to match the existing pattern/formatting that I use for `kSaveStateRev` elsewhere? otherwise I think this looks good, I'm building to test it now. there are some UI tweaks that I'll want to do, but those can happen later

awwbees:
oh right... I can just do that myself!

PR #282, opened 2021-09-27
Drop JuceHeader.h, include JUCE modules directly

mvf:
This is a first step to reduce build time and namespace pollution. It's quite mechanical:

- Remove `#include <JuceHeader.h>`
- `#include` the JUCE module header(s) only
- Add `juce::` / `using` as needed (`JuceHeader.h` was configured to do a global `using namespace juce`).

Speedup is rather modest, ~10% on Linux. If and when this gets merged, I have a second commit ready that makes the most notorious headers JUCE-free, reducing build time by 40%-50% vs. current `main`. Those changes move code around and change interfaces, so I separated them out for easier review.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,14 @@ project(BespokeSynth VERSION 1.0.1 LANGUAGES C CXX ASM)
 
 message(STATUS "Bespoke: Build Type: ${CMAKE_BUILD_TYPE}")
 
+# Global compile options
+add_compile_options(
+    # Build with Multiple Processes on Visual Studio
+    $<$<CXX_COMPILER_ID:MSVC>:/MP>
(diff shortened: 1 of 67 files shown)


baconpaul:
Wow this pr builds windows ci done in 13 minutes vs 22 for my input bus pr earlier in the day. Amazing. And you have another 50% coming? Ha!

baconpaul:
just did a quick eyeball. inasmuch as it is useful, other than the new conflict which looks pretty easy to resolve, the changes read correctly to me (namely you just added juce:: and added the appropriate header). And the azure change is smart yeah. If we don't have that in the surge azpipeline we should fix that!

awwbees:
wow, I hope you had a tool to assist in adding all those namespace qualifiers! I could feel my RSI tingling just by reading through those...

thank you so much!!!

PR #293, opened 2021-09-28
FileStream, ModularSynth: De-JUCE interface

mvf:
Makes `FileStream.h` and `ModularSynth.h` includible sans JUCE, dropping `juce_core.h` from 1st to 5th most expensive header for another modest compile time win.


Source/FileStream.cpp
```diff
@@ -25,114 +25,115 @@
 
 #include "FileStream.h"
 #include "ModularSynth.h"
-#include "SynthGlobals.h"
+
+#include "juce_core/juce_core.h"
 
 FileStreamOut::FileStreamOut(const char* file)
-: mStream(juce::File(file))
+: mStream(new juce::FileOutputStream{juce::File{file}})
```
awwbees: does this ever get freed?

awwbees (commented):
looks good other than these two news that I am unclear on. maybe there's a mechanism somewhere which deletes them that I'm not seeing?

Source/FileStream.cpp
mvf: It gets implicitly released by `mStream`'s dtor since it's a [`std::unique_ptr`](https://en.cppreference.com/w/cpp/memory/unique_ptr).

This is still a good finding. I'll change these to `std::make_unique`. Turns out it's already used elsewhere in the code, so Bespoke is already de-facto C++14. This will make things clear without having to look up `mStream`'s type. Thanks!

awwbees:
thanks!!

PR #297, opened 2021-09-29
`FileStream`: Fix non-ASCII paths

mvf:
With such paths, loading would crash and saving would silently fail or create files with a garbage name.

`juce::File*Stream` ctors take a `juce::File`, whose ctor in turn takes a `juce::String`. When constructing a `juce::String` from `std::string`, input is assumed to be UTF-8. When constructing from `char*`, it's assumed to be 7-bit ASCII. The encoding can be specified via `juce::CharPointer_UTF8` & friends, but that's cumbersome and we don't want complete JUCE types at the `FileStream` interface.

So this simply deletes the `char*` ctor overloads and only accepts `std::string`s. These are UTF-8 at all current call sites because they come from `juce::File`s.

### DIFF
--- a/Source/FileStream.cpp
+++ b/Source/FileStream.cpp
@@ -28,7 +28,7 @@
 
 #include "juce_core/juce_core.h"
 
-FileStreamOut::FileStreamOut(const char* file)
+FileStreamOut::FileStreamOut(const std::string& file)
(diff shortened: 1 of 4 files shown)


awwbees:
thank you!! yeah, I have so many `char*`s that I should refactor into `string`s, but have never gotten around to it...

Issue #299, opened 2021-09-29
Two crashes on livestreaming, typically doing lots of stuff with multiple sequencers and connections all over everywhere


airwindows:
backtrace:
0   BespokeSynth                        0x000000010e030400 _ZN4juce11SystemStats17getStackBacktraceEv + 64
1   BespokeSynth                        0x000000010df558fe _ZN12ModularSynth9DumpStatsEbPv + 510
2   BespokeSynth                        0x000000010e03061d _ZN4juceL11handleCrashEi + 13
3   libsystem_platform.dylib            0x00007fff7da53b5d _sigtramp + 29
4   libsystem_c.dylib                   0x00007fff7d8f902e __vfprintf + 16688
5   BespokeSynth                        0x000000010dd7f3c9 _ZN17NoteStepSequencer10DrawModuleEv + 809
6   BespokeSynth                        0x000000010df338e9 _ZN15IDrawableModule9DrawFrameEffbRfS0_ + 2057
7   BespokeSynth                        0x000000010df33c7d _ZN15IDrawableModule6RenderEv + 173
8   BespokeSynth                        0x000000010df6c1b9 _ZN15ModuleContainer4DrawEv + 89
9   BespokeSynth                        0x000000010ddabdd0 _ZN6Prefab10DrawModuleEv + 288
10  BespokeSynth                        0x000000010df338e9 _ZN15IDrawableModule9DrawFrameEffbRfS0_ + 2057
11  BespokeSynth                        0x000000010df33c7d _ZN15IDrawableModule6RenderEv + 173
12  BespokeSynth                        0x000000010df6c1b9 _ZN15ModuleContainer4DrawEv + 89
13  BespokeSynth                        0x000000010df595da _ZN12ModularSynth4DrawEPv + 890
14  BespokeSynth                        0x000000010df4c6ed _ZN20MainContentComponent6renderEv + 461
15  BespokeSynth                        0x000000010e44c75c _ZN4juce13OpenGLContext11CachedImage11renderFrameEv + 284
16  BespokeSynth                        0x000000010e44b341 _ZN4juce13OpenGLContext11CachedImage6runJobEv + 177
17  BespokeSynth                        0x000000010e04883b _ZN4juce10ThreadPool10runNextJobERNS0_16ThreadPoolThreadE + 59
18  BespokeSynth                        0x000000010e08fe99 _ZN4juce10ThreadPool16ThreadPoolThread3runEv + 73
[...]

PR #300, opened 2021-09-30
299 Two crashes on livestreaming, typically doing lots of stuff with multiple sequencers and connections all over everywhere

awwbees:
this solution was to double-buffer the list of pitches. I could reproduce the issue pretty quickly before this fix, and no longer can


awwbees:
fixes #299

PR #302, opened 2021-09-30
Load VST by name

KottV:
Hi,
This is a very basic proof-of-concept
1. Plugins are loading by it's name instead of file name.
~~2. Saved states are broken, fails on validate json.~~ need more testing
3. Need to associate name with unique id. There are often different plugins installed with the same name: VST3 and VST2, different versions in different paths.


KottV:
wow, occasionally savestate is working, o_O
i'll do more test

PR #303, opened 2021-09-30
minimap component and settings integration

fredericlb:
- adds an optional minimap to ease navigation in big projects. 
- clicking on it moves the current viewport to the clicked position
- the minimap can be toggled through user settings

### DIFF
--- a/Source/ModularSynth.cpp
+++ b/Source/ModularSynth.cpp
@@ -1,6 +1,7 @@
 #include "ModularSynth.h"
 #include "IAudioSource.h"
 #include "IAudioEffect.h"
+#include "OpenFrameworksPort.h"
 #include "SynthGlobals.h"
 #include "Scale.h"
 #include "Transport.h"
@@ -8,6 +9,7 @@
 #include "InputChannel.h"
 #include "OutputChannel.h"
 #include "TitleBar.h"
+#include "Minimap.h"
 #include "LFOController.h"
 #include "MidiController.h"
 #include "ChaosEngine.h"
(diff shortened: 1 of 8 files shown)


Source/ModularSynth.cpp
```diff
@@ -1880,7 +1882,20 @@ void ModularSynth::ResetLayout()
    titleBar->SetModuleFactory(&mModuleFactory);
    titleBar->Init();
    mUILayerModuleContainer.AddModule(titleBar);
-   
+
+   if (!GetUserPrefs()["show_minimap"].isNull() && GetUserPrefs()["show_minimap"].asBool()) 
+   {
+      delete mMinimap;
```
cassaundra: Line 1888 invokes undefined behavior since `mMinimap` may not be allocated, causing a crash on some systems. This should be a simple fix, by setting it to `nullptr` when it is unallocated and checking against that.

cassaundra:
Other than the crash at startup and the committed VS code configuration file, this looks fantastic to me! It's useful but unobtrusive, and looks aesthetically in line with the rest of the program.

Source/ModularSynth.cpp
fredericlb: Thanks for the review ! It should be fixed now

fredericlb:
> Other than the crash at startup and the committed VS code configuration file, this looks fantastic to me! It's useful but unobtrusive, and looks aesthetically in line with the rest of the program.

I squashed my commits to remove the vscode file, should be ok now

awwbees:
sorry that it took me so long to look at this one. this is super slick, thanks! I've got a few tweaks I want to make brewing in my head (dragging around within the minimap, and seeing if I can preserve aspect ratio in a way that feels good), but this is a really solid start.

PR #304, opened 2021-09-30
clang-format integration

fredericlb:
working on multiple projects with different coding styles I use clang-format to automatically apply formatting rules on save. 
i know some devs don't like this kind of tools, so it's just a suggestion. 

i tried to adapt rules as close as possible to current project coding style, but of course on already existing files it makes some changes (i do not use it on files not created by me). clang-format has good support in visual studio code and clion and seems to be also supported by visual studio and xcode.


mvf:
Very much in favor of `clang-format`. The style, once finalized, should be applied to the full code base (also fixing the inconsistent `LF`/`CRLF`). Then it can be enforced in the pipelines so things stay tidy. That's what we do in Surge, and personally I wouldn't want to go back.

baconpaul:
100% agree.

awwbees:
sweet, thank you so much. this is my first time seeing clang-format, so I will learn all the details of these config options and update them to match the style that I've been using

baconpaul:
https://zed0.co.uk/clang-format-configurator/

that is super useful if you want to understand the options

awwbees:
I finally did this in #808, thank you for pointing me towards this!

PR #306, opened 2021-09-30
Fix most deprecation warnings

mvf:
These were starting to get annoying. The only remaining deprecation warning is in the `libusb` macOS implementation. It will be addressed in a separate PR that deals with `push2`.

---

- Migrate away from deprecated C++, JUCE and libusb API
- Remove `-Wrange-loop-analysis` pragma (GCC does not implement this warning, so it warns about the pragma itself)

### DIFF
--- a/Source/ControllingSong.cpp
+++ b/Source/ControllingSong.cpp
@@ -89,7 +89,7 @@ void ControllingSong::Init()
       mSongSelector->AddLabel(title.c_str(), i);
    }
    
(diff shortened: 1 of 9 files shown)


awwbees:
ooh, some nice fancy new (to me) syntax in here. good stuff for me to learn from!

mvf:
This JUCE file API makes me all kinds of jealous too. In Surge we're using plain `std::filesystem` and it's pretty barebones.

PR #307, opened 2021-09-30
Set language level to C++17

mvf:
So far we didn't specify a C++ standard. This works when compiling everything in the main executable where JUCE sets it to C++14. But breaking the project into libraries causes them to be built with the compiler default. On AppleClang that's C++98, so the build fails.

C++17 packs many useful features and is well-supported these days. The only slightly troubled platform in this regard is macOS.

This change might raise the minimum build requirement on macOS to Xcode 9.3, which seems to require macOS 10.13 "High Sierra". The deployment target (i.e. minimum macOS to run the Bespoke binary) remains unchanged at macOS 10.11 "El Capitan".


mvf:
Okies, so turns out Bespoke currently only builds on Windows more or less by accident. I'll prepare another PR to rectify this first.

mvf:
Merging #308 would fix this PR.

PR #308, opened 2021-09-30
Purge "using namespace std" from the code base

mvf:
When setting the C++ standard to C++17, all hell breaks loose on Windows due to ambiguity between `std` and the global namespace. (see #307)

This change removes `using namespace std` and tacks `std::` in front of everything that needs it. No more, no less.


mvf:
A handy command line to ~vet this for rickroll comments~ review this:
```
$ git show --color-words --unified=0
```

awwbees:
this one is gonna take some getting used to!!

mvf:
It does add a bit of clutter, but IME it's worth it.

This deliberately brain-dead change also makes it look somewhat worse than it is. These days `auto` is often enough. I didn't use it here because that would have taken much longer and would have been harder to review.

PR #311, opened 2021-10-01
Use Enabled checkbox for prefabs

zauner:
This allows enabling/disabling all modules within a prefab with one checkbox on prefab level.

More like a proof of concept, as I don't know about any possible side effects, and if this has been discussed and dismissed already.

### DIFF
--- a/Source/Prefab.cpp
+++ b/Source/Prefab.cpp
@@ -233,6 +233,17 @@ void Prefab::ButtonClicked(ClickButton* button)
    }
 }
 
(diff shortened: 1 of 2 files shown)


awwbees:
this one is a little dicey: it would "forget" which modules are enabled disabled within the prefab. I might add this feature, but I will do it in a way the doesn't override the module's independent enabled state.

zauner:
Yes, I see your point -- so making `Enabled()` recursively consider the parent's state would be a more robust solution.

Anyway, I hope this feature will find its way into a release, and let you do your thing :)

PR #313, opened 2021-10-01
Add mapping for Akai APC Mini controller

IrregularShed:
I've got an APC Mini lying around doing nothing now that I've got some bigger, flashier
equipment. Rather than leave it lying around I thought I'd put it to good use.

image


awwbees:
thanks!!

Issue #133, opened 2021-09-16
Trying to use scripting crashes on Mac OS 10.15.7


rbnpi:
Python3 installed with brew and set to default (so in terminal python --version =>Python 3.9.7) When scripting is selected, as soon as you try and tap anything in Bespoke quits.( EDIT I tried changing to python 3.8.12 Same problem)
Are there any specific setup procedures required for it to recognise python as installed?

awwbees:
are you seeing this warning? and then after you click it, it crashes?
![image](https://user-images.githubusercontent.com/839239/133718022-76965530-1c61-4ad2-bcb7-f2b4eead9aee.png)

it is expected that 3.9 would crash, but 3.8 should have worked

hmm. it sounds like maybe it's not finding python correctly. do you know where your python 3.8 is installed?

rbnpi:
Yes I do see that warning. I (now) have python@3.8 installed by Brew on my Mac. When I click to continue it quits.
I also have brew installed python@3.9 but this is unlinked.
I have the following links in place to access it
which python => /usr/local/bin/python
ls -l /usr/local/bin/python => /usr/local/bin/python3
ls -l /usr/local/bin/python3 => ../Cellar/python@3.8/3.8.12/bin/python3

i.e. its in /usr/local/Cellar/python@3.8/3.8.12/bin
This contains
2to3			pip3.8			python3.8
2to3-3.8		pydoc3			python3.8-config
idle3			pydoc3.8		wheel3
idle3.8			python3
pip3			python3-config


python --version => Python 3.8.12

joshlacey:
I'm seeing the same issue on Big Sur 11.1

python --version => Python 3.8.2

Tadaboody:
Am getting the same issue on big sur 11.6 and python3.9 installed globally when running bespoke through the terminal this was the crash dump
```
❯ ./Contents/MacOS/Bespoke
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
Python path configuration:
  PYTHONHOME = (not set)
  PYTHONPATH = (not set)
  program name = 'python3'
  isolated = 0
  environment = 1
  user site = 1
  import site = 1
  sys._base_executable = '/Applications/Bespoke.app/./Contents/MacOS/Bespoke'
  sys.base_prefix = '/Library/Frameworks/Python.framework/Versions/3.8'
[...]

PR #314, opened 2021-10-01
Fix crash when first time user opens .bsk with scripts.

diimdeep:
More info https://github.com/BespokeSynth/BespokeSynth/issues/133#issuecomment-932276570

### DIFF
--- a/Source/ScriptModule.cpp
+++ b/Source/ScriptModule.cpp
@@ -90,7 +90,7 @@ ScriptModule::ScriptModule()
 , mShowJediWarning(false)
 {
    CheckIfPythonEverSuccessfullyInitialized();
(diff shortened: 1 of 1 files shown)


baconpaul:
So @awwbees @diimdeep and i were chatting about this on discord and comparing with #315 

I think this PR has the effect of "if you load a BSK file and have never loaded python it does the equivalent of clicking on start python". On a system with python this will mean you don't get a crash but on a system without python I think you would then try and init python and fail

#315 has the different effect of "if you load a BSK file and have never loaded python, the python doesn't show up and you get the click to start python window"

So basically if you blow away the python_installed and load a .bsk with ScriptModule in it

1. This PR would make it the same as if you clicked 'start python' without you clicking and will show the python
2. #315 will show you the 'click to start python' message without crashing and if you click it will start

So I think (think) #315 might be safer for non-python systems loading bsk files.

diimdeep:
@baconpaul Thanks, now I think this is not right change to do.

PR #315, opened 2021-10-01
Python Startup Crash Avoidance on .bsk load

### LINKED ISSUE #133: Trying to use scripting crashes on Mac OS 10.15.7

baconpaul:
In the case where

1. Your system has a valid python
2. But the python_ever_init file marker isn't there and
3. You directly load a BSK with python that was saved from a working session

you would crash. You would crash because the syntax highlighter would run.
You can repro this by save a BSK with python, blow away the sentinel, and then
restart bespoke and load it.

This fixes the problem by

1. Making the code highlight sentinel be static and default to false
   and toggle to true when py is initialized
2. Keep a per-module instance highlighter to unstream onto to preserve
   backwarad capability

Closes #133

### DIFF
--- a/Source/CodeEntry.cpp
+++ b/Source/CodeEntry.cpp
@@ -50,7 +50,8 @@ namespace py = pybind11;
 
 //static
 bool CodeEntry::sWarnJediNotInstalled = false;
-
+bool CodeEntry::sDoPythonAutocomplete = false;
+bool CodeEntry::sDoSyntaxHighlighting = false;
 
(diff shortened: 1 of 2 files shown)


awwbees:
awesome

PR #319, opened 2021-10-01
push2: Lib-ify

mvf:
- Add libs/CMakeLists.txt
- Move push2 to libs
- Hide libusb dependency from rest of codebase
- Linux/BSD: Fix header mismatch (i.e. including our libusb.h but
  linking against system libusb)
- Use system libusb on MinGW


baconpaul:
Nice!

awwbees:
oh interesting, cool thanks! so I'm guessing this makes it easier to build for folks who don't care about push 2 support? (which is the vast majority of users)

awwbees:
it's definitely interesting to know about this technique for other obscure dependences like SpaceMouseControl (of which I will probably be the only client)

mvf:
> so I'm guessing this makes it easier to build for folks who don't care about push 2 support?

This was just the groundwork, #322 makes Push support optional with graceful fallback on missing `libusb`.

PR #322, opened 2021-10-02
Make Ableton Push 2 support optional

mvf:
This adds the `BESPOKE_PUSH2_SUPPORT` build option (defaults to `ON`). When disabled, Bespoke builds without `libusb` and the `push2control` module shows a "no push 2 support" message.

image

On anything not macOS or Visual Studio we use the system `libusb`. If it can't be found, warn and gracefully disable Push 2 support.

Also fixes leaking of the Push API objects at shutdown and on every click on a disconnected `push2control`.

### DIFF
--- a/Source/Push2Control.cpp
+++ b/Source/Push2Control.cpp
@@ -49,13 +49,16 @@ using namespace juce::gl;
 #include "DropdownList.h"
 #include "FloatSliderLFOControl.h"
 #include "UserPrefsEditor.h"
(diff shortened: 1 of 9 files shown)


awwbees:
awesome!! and I just tested with my push2, still works fine

mvf:
Many thanks for testing! I don't have a Push, so there's only so much synthetic testing under the debugger I can do.

PR #323, opened 2021-10-02
3DxWare: Lib-ify

mvf:
Moves the 3DxWare library to `libs` and adds a `BESPOKE_SPACEMOUSE_SUPPORT` build option. Can only be enabled on Visual Studio and is enabled there by default.


awwbees:
man, I am really loving how easy this new cmake set up makes it to libify things. tested with my spacemouse to confirm that it went off without a hitch.

mvf:
Sweet!

PR #325, opened 2021-10-02
Fix MinGW builds

mvf:
This one's for my homie @NoxyNixie.

---

This fixes the trivialities that make MinGW builds fail. The sysroot still needs Python, but that's a non-issue when building natively.


awwbees:
looks like some really subtle and annoying fixes!

PR #326, opened 2021-10-02
Make it possible to rescan VSTs with a huge list of VSTs.

NoxyNixie:
Changed the 'rescan vsts' menu entry to be in the front of the list instead of behind potentially hundreds of VSTs.

(I am aware that a better UI solution is in the works but for the time being this seems to be a quick and easy "fix")


awwbees:
I have some VST scanning stuff currently in the works, so I'll hold off on this one.

PR #329, opened 2021-10-03
exprtk: Lib-ify

mvf:


awwbees:
lib-ifying the simplest lib in the world. I love exprtk.

PR #330, opened 2021-10-03
psmove: Lib-ify

mvf:
Similar treatment as for Push 2: The `psmove` module is now available everywhere, but it's only functional where we have a HID backend.


mvf:
Note: Other than `push2control`, the `psmove` module won't indicate when it's a dummy. I hope that's cool, the module isn't even exposed to users yet. I actually have it building for both Windows and Linux but wanted to keep this PR nice and focused. Also, I own a PS Move and would rather do some testing before shipping support for new platforms.

awwbees:
nice! I don't currently have hardware setup to test this, and haven't used a PS Move with bespoke in at least 6 years or so, but I'm happy to trust that it's at least not unfixably broken by this change. it is a hidden module, as you said, so basically as long as it's passed those compile checks, I'm happy.

PR #334, opened 2021-10-04
VST Midi Output in the VSTPlugin module

baconpaul:
This adds a working VST Output point for VST Midi Output
plugins. There's a couple of problems

1. Independetly of this, it seems tempo and position in VST is
   wrong so stochas (my test case) acts a bit wierd
2. I positioned the midi output on the plugin but we could put it
   elsewhere too. Wasn't sure where to put it. I thoughts
   also about putting it in the bottom of the module maybe?
3. I couldn't figure out how to label the midi output either.

So I'm not entirely sure this is something you just want to merge
rather than sharing comments on 2 and 3, but also i don't think it
is broken and is mergable.

### DIFF
--- a/Source/VSTPlugin.cpp
+++ b/Source/VSTPlugin.cpp
@@ -550,13 +550,14 @@ void VSTPlugin::Process(double time)
             while (midiIt != mMidiBuffer.end())
             {
                auto msg = (*midiIt).getMessage();
+               auto tMidi = time + (*midiIt).samplePosition * gInvSampleRateMs;
                if (msg.isNoteOn())
                {
-                  mMidiOutCable->PlayNoteOutput(time, msg.getNoteNumber(), msg.getVelocity());
+                  mMidiOutCable->PlayNoteOutput(tMidi, msg.getNoteNumber(), msg.getVelocity());
                }
                else if (msg.isNoteOff())
                {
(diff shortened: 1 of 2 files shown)


Source/VSTPlugin.cpp
```diff
@@ -535,7 +543,30 @@ void VSTPlugin::Process(double time)
...
+
+         if (!mMidiBuffer.isEmpty())
+         {
+            auto midiIt = mMidiBuffer.begin();
+            while (midiIt != mMidiBuffer.end())
+            {
+               auto msg = (*midiIt).getMessage();
+               if (msg.isNoteOn())
+               {
+                  mMidiOutCable->PlayNoteOutput(time, msg.getNoteNumber(), msg.getVelocity());
```
awwbees: I think this `time` (and the one on line 559) needs an offset added based upon the timestamp; with this code, the note playback times will be quantized to the buffer size. I can take a look at this later if you don't want to dig in to this side of it.

awwbees (changes requested):

Source/VSTPlugin.cpp
baconpaul: You are correct! Hold on lemme peek.

baconpaul: OK pushed a change that does this

awwbees (approved):

PR #339, opened 2021-10-04
Split CMakeLists.txt, remove obsolete .jucer, glew

mvf:
Move the Source-relative stuff (i.e. the BespokeSynth binary) to its own CMakeLists.txt so we don't have to repeat "Source" everywhere.

Make the global BESPOKE_* build options cache variables for consistency and easy discovery in the CMake GUI.

Remove the .jucer file and GLEW library (both obsolete with JUCE 6)


baconpaul:
looks good to me!

awwbees:
oh excellent. I was just looking at that jucer yesterday and thinking of deleting it.

PR #341, opened 2021-10-05
 Add the option for displaying native window decorations for plugin's     window

KottV:
This allows to use native titlebar. So it's possible to toggle "stay on top" or "omnipresent" for plugin's window.
изображение


awwbees:
do you think there's ever a reason that a user would not want to use the native title bar? maybe we should just enable the native title bar, and not clutter up the settings with an unnecessary checkbox.

KottV:
> do you think there's ever a reason that a user would not want to use the native title bar? maybe we should just enable the native title bar, and not clutter up the settings with an unnecessary checkbox.

I have a very humble assumption that it may be a case for the Wayland on Linux (where the server-side decorations may not work), otherwise - not.
But I haven't test it yet in Wayland. We can pend it till I test it for such case.

KottV:
I removed the option, if it cause some drawbacks then we always can re-add it.

KottV:
@awwbees is it ok to merge now?

awwbees:
looks good!

awwbees:
this feature is causing me to crash when opening the plugin window for Vital in Windows, I'm going to undo this change.

it also didn't enable any "always on top" functionality for me in Windows, the native window in Windows doesn't seem to include that. I'm going to instead add a checkbox to `vstplugin` for "always on top", and use non-native windows.

PR #343, opened 2021-10-05
Generic event propagation infrastructure.

KaruroChori:
This is the first commit to detach keyboard events from assigned ASCII characters, but also to support other types of events in the future.
Keycodes are way better to control instruments, as we can map arbitrary macros from multiple input devices without interfering with the actual control keys for the interface. Macropads, mechanical keyboards, gamepads, they could all be used to interface with our virtual tools.
I implemented a test for the KeyboardDisplay which was a bit broken in how it handled input events in my opinion due to the overlay of actions on the interface and events on the KeyboardDisplay itself.
Next step is to add configurable maps and remove the hardcoded one.

The questionable solution based on the pair const void*, size_t is meant to be made more ideomatic in the future, once the Event class will be fully developed.

### DIFF
--- a/Source/Event.h
+++ b/Source/Event.h
@@ -1,0 +1,7 @@
+#pragma once
+
+#include "juce_opengl/juce_opengl.h"
+
+struct Event{
+    static inline const juce::KeyPress& keycode(const void* data, size_t size){return *(const juce::KeyPress*)data;}
+};
--- a/Source/IDrawableModule.cpp
+++ b/Source/IDrawableModule.cpp
@@ -796,6 +796,12 @@ ofVec2f IDrawableModule::GetMinimumDimensions()
    return ofVec2f(GetMinimizedWidth() + 10, 10);
 }
 
+void IDrawableModule::SignalEmit(const void *data, size_t size)
+{
+   for (auto control : mUIControls)
+      control->SignalEmit(data, size);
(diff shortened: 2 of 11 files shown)


awwbees:
abstracting keypresses into an Event class is a great idea, it will offer a lot of flexibility. however, I think this change is a little too raw to pull into bespoke at this point, I'm not comfortable passing around void pointers at such a high level. this is too work-in-progress to merge.

I have a refactor planned to allow for custom keybindings, and I will keep this Event design in mind while doing it. I think it's a good one; it will allow for very flexible binding, as you say.

KaruroChori:
Agreed, this was just a prototype, it is not really meant to be merged at this stage. It was more to receive feedback over the proposal and to harmonize it with the other developments.
That void pointer is there as I started writing the propagation backwards, so I had no idea which types would have been involved reaching the root :D.

KaruroChori:
No more void pointers :D.
Also, I started working on the generation of UI events starting from Keyboard events. The idea would be to have a virtual component which is going to capture all keyboard events and generate keycode-independent ui events to be propagated to the rest of the dependency tree. This would allow ui-events to be simulated and triggered by any other kind of input device.

Is there a list of UI events easily accessible somewhere in the documentation or in the code?

PR #344, opened 2021-10-05
More VST Bus Management

baconpaul:
This commit makes Surge 1.9 work in Bespoke as well as Surge XT.
It works around the case when synths (of which Surge 1.9 is one, but
Surge XT is not) cannot handle disabled or alternate bus configurations
gracefully. It has three big changes

1. Call disableNonMain Buses on the VST object
2. Remove the 'MAX' for mOutput and mInput which was arbitrarily
   set at 4
3. In the event that after you disable those buses you still
   need more than 2 outputs, allow that to occur by including
   the max of inputs and outputs in the buffer allocation in
   process

### DIFF
--- a/Source/VSTPlugin.cpp
+++ b/Source/VSTPlugin.cpp
@@ -349,23 +349,27 @@ void VSTPlugin::LoadVST(juce::PluginDescription desc)
    if (mPlugin != nullptr)
    {
       mPlugin->enableAllBuses();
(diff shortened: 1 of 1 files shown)


baconpaul:
Oh hold up - small change to this inbound

baconpaul:
OK force pushed a corrected version. This one is now ready for review. Sorry bout that!

awwbees:
thanks!!

Issue #30, opened 2020-08-21
[Linux] BespokeSynth only works well when compiling in Debug mode.


xunil-cloud:
I try to compile from source on Arch Linux. If I compile in **Release** mode, the computer keyboard is not really responsive. For instance, the popup menu won't show up every time when I press 'k'. Another wired thing is that BespokeSynth has no sound in JACK and only has sound in pulseAudio. I also get some messages from terminal:  

```
44.8387: trying to stop before we started (44845<=44845)
45.5121: trying to stop before we started (45517.6<=45517.6)
47.5322: trying to stop before we started (47535.4<=47535.4)
48.2056: trying to stop before we started (48208<=48208)
```
I also try the latest [v0.0004-pre](https://github.com/awwbees/BespokeSynth/releases/tag/v0.0004-pre) binary version, but it has the same issues.

Note: If I compile in **Debug** mode, I have no issue. Extremely AWESOME!

djdeath:
Kind of feels similar to https://github.com/awwbees/BespokeSynth/issues/29

Another thing I noticed for instance is that trying to record the issue with the gnome-shell screen record makes it go away.
I wonder if there is some kind of race condition in the event processing code that would make this happen.
Since you mentioned that a Debug build has no issue, it might just slow things enough to avoid the race... :man_shrugging:

xunil-cloud:
Did you try the keyboarddisplay module? I have similar mouse click issue that you have just described in #29. By the way, I also try a Release build without optimization, but that doesn't fix the problem.

awwbees:
that is very strange! maybe I should release a debug build for linux until I figure it out

xunil-cloud:
I am also trying to figure out why this happened. If I find something new, I will let you know. 🙂

awwbees:
@xunil-cloud if in https://github.com/awwbees/BespokeSynth/blob/master/Source/MainComponent.cpp, you change:
```   static int sRenderFrame = 0;
#if DEBUG
      if (sRenderFrame % 2 == 0)
#else
      if (true)
#endif
      {
         openGLContext.triggerRepaint();
      }
```

to simply

```   static int sRenderFrame = 0;
      if (sRenderFrame % 2 == 0)
      {
         openGLContext.triggerRepaint();
      }
```

does that make it perform like debug? the side effect is that this will limit the display frame rate to 30fps

djdeath:
[...]

PR #349, opened 2021-10-06
Fixes performance issues on Linux, Run @60fps

### LINKED ISSUE #309: Framerate locked at 32 fps on Linux
For some reason Bespoke runs extremely poor for me on X11/Openbox with or without compositor on intel igpu. Fps is locked at 32

pushqrdx:
This should fix #309 which originated from #30. Original issue seems to be caused by how JUCE sends batch render calls to the gpu causing hangs with mesa/dri leading to dropped input events, after fiddling around a bit i found that if you're handling rendering yourself (not using the paint override) like Bespoke does, you should set `ComponentPaintingEnabled` to false to improve performance and avoid redundant call to paint. 
Note: this might also benefit performance on other platforms.

From Juce Docs:
> By default this is set to true. If you're not using any paint() method functionality and are doing all your rendering in an OpenGLRenderer, you should disable it to improve performance.

---

This fixes the issue on my machine and Bespoke runs at 60fps with no issues.

### DIFF
--- a/Source/MainComponent.cpp
+++ b/Source/MainComponent.cpp
@@ -52,6 +52,7 @@ public:
       
       openGLContext.setOpenGLVersionRequired(juce::OpenGLContext::openGL3_2);
       openGLContext.setContinuousRepainting(false);
+      openGLContext.setComponentPaintingEnabled(false);
 
 #ifndef BESPOKE_WINDOWS //windows crash handler is set up in ModularSynth() constructor
       SystemStats::setApplicationCrashHandler(ModularSynth::CrashHandler);
@@ -121,7 +122,7 @@ public:
       
       mSynth.Poll();
       
-#if DEBUG || BESPOKE_LINUX
+#if DEBUG
       if (sRenderFrame % 2 == 0)
 #else
       if (true)


cassaundra:
I'm glad to see this issue being address! Bespoke feels much smoother when at 60 fps. Unfortunately, I still seem to be getting the same input issues as before, on X11/Mesa.

Source/MainComponent.cpp
```diff
@@ -121,11 +122,7 @@ class MainContentComponent   : public OpenGLAppComponent,
       
       mSynth.Poll();
       
-#if DEBUG || BESPOKE_LINUX
```
awwbees, line 124: could you leave the DEBUG check in, and just remove BESPOKE_LINUX here? I still want to run at 30fps in debug mode

cassaundra:
Aha, I was wrong: it turns out this fix does work. After an hour of troubleshooting and digging through JUCE internals, it turns out I was just executing an old binary. It seems the release directory was moved from `./ignore/build/BespokeSynth_artefacts/Release` to `./ignore/build/Source/BespokeSynth_artefacts/Release` in 1f68e48b765430d65b46e21b93cd22d4eef540e0, although I'm not sure if this was an intentional side effect.

Source/MainComponent.cpp
pushqrdx: @awwbees sorry forgot to add it back :D

awwbees:
amazing!! thank you!!!!

awwbees:
I unfortunately had to temporarily undo this change, it is leading to thread safety crashes. it seems that I have some thread-unsafe code in my rendering, and the fix in this PR uncovered these issues. I need to fix these issues, so I can restore the fix in this PR, but until I do that, I need to undo this PR

if you still wish to run with this change (and risk random crashes), you can re-enable the behavior by generating your cmake project with `-DBESPOKE_LINUX_HIGH_FPS_WITH_THREAD_PROBLEMS:BOOL=TRUE`

PR #351, opened 2021-10-06
Event propagation

KaruroChori:
First working prototype of the revised event propagation, with 100% less void pointers. And this time I even found the button to tag the PR as WIP.
It is built alongside the current system for the time being, but my plan is to rewrite all the components to make use of the new one in the future.
At this stage it can propagate different signals for the raising and falling edge of a key being pressed as well as its sustained state.
There are also some placeholders for what is next, signal layers (to handle their visibility in bulk) and matching functions to generate new signals from macros of others.

Any feedback, even more so on the architectural side would be appreciated :)


KaruroChori:
I cannot replicate these building errors on my system. I will have to look into those :disappointed: .

awwbees:
sorry... I have a lot of trouble following this code, it seems that we must think in very different paradigms.

I think making a generic events system is a great idea, which I will pursue in the future. I would prefer to write it in a way that makes more intuitive sense to me, though, and I don't see a way to "steer" this code into being closer to that. sorry if this is disappointing!

KaruroChori:
I mean, you are clearly free to pursue this in the way which best suits you, and the same is true for me or anyone else after you released it as GPL :).
Sure, I was hoping for a more meaningful conversation on an architectural level than this, even just so that my actions could have been better coordinated with the long term vision you had for Bespoke; but honestly, as long as my code does work for me, there is not really much more I need; so no real disappointment on my side.

Setting my proposal aside, on a purely technical level I believe that the way the current event model is handled will be a nightmare in the future if not addressed sooner than later. It may pose a strong limitation on what can be done with your software and the capabilities of interfacing it with external devices. If not my design, some solution will need to be picked at some point.

Good luck!

PR #352, opened 2021-10-06
Allow multi patching of special cable types.

### LINKED ISSUE #331: multiple bugs on presets module
- it's not possible to connect multiple things to one `presets` module (the circle shows a + on hover, but dragging from it does nothing)
- it's not possible do disconnect existing cable from `presets` module or reroute it (you need to delete whole `presets` module)
- maybe we should have *save preset* button in addition to shift+click saving (this will be more obvious)

NoxyNixie:
Allow multi-patching of kConnectionType_UIControl and kConnectionType_Special cable types (partially fixes #331).

### DIFF
--- a/Source/PatchCableSource.cpp
+++ b/Source/PatchCableSource.cpp
@@ -84,6 +84,10 @@ void PatchCableSource::SetConnectionType(ConnectionType type)
       mColor = IDrawableModule::GetColor(kModuleType_Modulator);
    else if (mType == kConnectionType_Pulse)
       mColor = IDrawableModule::GetColor(kModuleType_Pulse);
+   else if (mType == kConnectionType_Special)
+      mColor = IDrawableModule::GetColor(kModuleType_Instrument);
+   else if (mType == kConnectionType_UIControl)
+      mColor = IDrawableModule::GetColor(kModuleType_Modulator);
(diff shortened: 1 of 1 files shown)


awwbees:
sorry for the delay in merging this one... I need to do some testing to make sure that this is correct in all scenarios

NoxyNixie:
I agree. I've been mucking around in the source code in various places (mostly to learn the project and hopefully fix bugs along the way) and I did find that the presets module is quite temperamental but is perfectly okay with these changes.

I've also tested the kConnectionType_Special in `prefab` and `loopergranulater` and those work fine. Though the `loopergranulater` has other issues with regards to unpatching/repatching (the loopers stay in granulation mode even when disconnected, you have to delete the original looper granulator for it to clear up but I suspect that is a completely different issue and not related to this change at all since that has happened even before this patch).

For the kConnectionType_UIControl I tested with `eventcanvas`, `midicontroller`, `valustream` and lfo.

Aside from the seemingly unrelated thread crash issue (This crash happens regardless of this change (and only in Debug mode)) I have not had any issues with this change.

awwbees:
I ended up making nearly identical changes in my #546 PR, so this is all set now!

PR #357, opened 2021-10-07
Add mapping for the knobs across the top of an alesis vmini

baconpaul:
https://www.alesis.com/products/view/vmini

I have one of those at my dev desk. Super handy. Here's a mapping for you to take if you want or nuke if you don't! :)


awwbees:
I accept all controller mapping files!

PR #358, opened 2021-10-08
Add tooltip translation (german)

ghost:


awwbees:
wow! thank you!!! it appears that you have left a .gitignore in this change accidentally... could you remove that part? and then I will happily merge!

ghost:
Done. (It wasn't accidental but I removed it now)

PR #359, opened 2021-10-08
build a hard limiter into the output module, and worn users more explicitly when they're clipping

awwbees:


awwbees:
*warn. stupid speech-to-text!

PR #362, opened 2021-10-08
Fix and simplify resource directory discovery

### LINKED ISSUE #354: "Couldn't load font" error when opening bsk on Windows via double click
I haven't been able to fully figure out where this is coming from but if a patch is saved on a mac os copy of Bespoke and then you try and open in on a windows machine by double-clicking the file you get the error:

"Couldn't load font from C:/Users/USERNAME/MacOSX/build/Release/resource/frabk.tff"

Where USERNAME is the username you are logged into. Strangely, opening up Bespoke and loading the same file from the "Load" button seems to work fine.

I thought maybe it was just me but I have seen at least one other person on the discord run into the same inssue.
Screenshot 2021-10-06 194714

mvf:
Follow-up of a Discord chat with @baconpaul today. Fixes #354

---
The resource directory was resolved relative to the CWD everywhere
except on macOS, causing resources to not be found in many cases.

This removes reliance on the CWD and only looks in one well-defined path
(two on non-Apple UNIX). If the resource dir ain't there, that's a usage
error and we throw.

The paths are as follows:

macOS: \<Bundle directory\>/Contents/Resources/resource
Windows: <BespokeSynth.exe directory>/resource
UNIX: \<BespokeSynth executable directory\>/resource
      \<prefix\>/share/BespokeSynth/resource (for distros)

### DIFF
--- a/Source/BeatBloks.cpp
+++ b/Source/BeatBloks.cpp
@@ -311,7 +311,7 @@ void BeatBloks::FilesDropped(std::vector<std::string> files, int x, int y)
    if (!hasCached)   //have to look it up with echonest
    {
       char command[2048];
-      sprintf(command,"export ECHO_NEST_API_KEY=SUZ3W7PAIVQQXZCAW; export PATH=/usr/local/bin:$PATH; python \"%s\" \"%s\"", ofToDataPath("get_echonest_remix_data.py",true).c_str(), files[0].c_str());
+      sprintf(command,"export ECHO_NEST_API_KEY=SUZ3W7PAIVQQXZCAW; export PATH=/usr/local/bin:$PATH; python \"%s\" \"%s\"", ofToDataPath("get_echonest_remix_data.py").c_str(), files[0].c_str());
       output = popen(command, "r");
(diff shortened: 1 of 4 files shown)


awwbees:
ahh... so much cleaner!

baconpaul:
My only comment: The `throw std::runtime_error` - is that caught and turned into a user error message or will a user just get a crash to desktop if misconfigured?

a zillion times better tho yeah!

mvf:
> The `throw std::runtime_error` - is that caught and turned into a user error message or will a user just get a crash to desktop if misconfigured?

That's actually a very good point. The exception is not caught and we'll crash. This prints the error to the terminal, but a Windows user might be lost. And this only throws when the folder isn't there. If there's a file missing inside, the behavior will be as before. I should change that to something else, blank buttons are easier to diagnose than a crash. I just had my Python hat on where `ofToFactoryPath` will be called to locate the `python` folder, and if we get that wrong we'll crash anyway.

The alternative would be to install a top-level exception handler that shows a message box. This would allow us to crash gracefully in more situations. Any thoughts @awwbees?

PR #365, opened 2021-10-09
fix issue with Linux apparently crashing on some VSTs when requesting "always on top"

awwbees:
new solution: Linux gets native title bars, other platforms get always on top

(note: vital was crashing on windows when I tried to open it with native title bars)


awwbees:
@KottV does this fix resolve your issue?

KottV:
> @KottV does this fix resolve your issue?

@awwbees 
yes, all is good
thank you! :)

KottV:
at least on Linux side

PR #366, opened 2021-10-09
Set BespokeSynth as the startup project for MSVC.

NoxyNixie:


awwbees:
oh excellent! I looked into this last week, but incorrectly concluded that we couldn't do this with the current cmake version

PR #369, opened 2021-10-09
Build a crude nightly asset on Win and Lin also

baconpaul:
This commit builds a crude nightly asset on windows and linux.


baconpaul:
Don't merge this one @awwbees - I'm gonna downgrade it to a draft since @mvf has a cleverer idea (which will probably also change the cmake part but not the azure part of my last one).

PR #372, opened 2021-10-10
Don't export `pybind11` symbols

mvf:
Slightly smaller binary, gets rid of the import lib and exports file on Windows:

image


baconpaul:
Oooh maybe we should add an azure step to make sure that .exp and .lib don't return at build time? Just thinking out loud.

mvf:
I have no opinion on adding a test for that. Just wanted to make sure there's no garbage in the output directory for when we later point the installer machinery at it. The `.exe.manifest` will also go away in installer builds since they link our own manifest into the binary.

PR #373, opened 2021-10-10
Add a scoped locale guard

baconpaul:
Force LC_NUMERIC to "C" for all json reads and writes.

### DIFF
--- a/Source/ofxJSONElement.cpp
+++ b/Source/ofxJSONElement.cpp
@@ -14,6 +14,31 @@
 
 using namespace Json;
 
+class ScopedLocale
+{
+public:
+    ScopedLocale() noexcept
+#if WINDOWS
+    : locale(::_strdup(::setlocale(LC_NUMERIC, nullptr)))
+#else
+            : locale(::strdup(::setlocale(LC_NUMERIC, nullptr)))
+#endif
+    {
+       ::setlocale(LC_NUMERIC, "C");
+    }
+
+    ~ScopedLocale() noexcept
+    {
+       if (locale != nullptr)
+       {
+          ::setlocale(LC_NUMERIC, locale);
+          ::free(locale);
+       }
+    }
+
+private:
+    char *const locale;
+};
 
 //--------------------------------------------------------------
 ofxJSONElement::ofxJSONElement(const Json::Value& v) : Value(v)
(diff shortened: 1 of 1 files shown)


baconpaul:
@awwbees I don't know if you want to merge this but I'm offline today so wanted to share. This is how we fixed the various numeric internationalization problems once and for all in surge - make a little guard class (which we got from falktx) to clobber the numeric LOCALE before all reads and writes.

I think this may  solve @KottV 's problem but we don't hav ea repro. Putting it here so you can see how we would do it since there's not exactly an issue to hang this idea on right now I don't think.

baconpaul:
OK so @awwbees we *should* merge this. Here's why

Your fix sets the local in SynthInit which is good
But in Kotts case a plugin resets the locale so blammo

I can reproduce a , in my json exactly by taking my tempo tester vst3 and adding to the ctor

```
std::locale::global(std::locale("es_ES"));
```

then when I load this VST in my bespoke, json files no longer save properly at head, but with this patch, they do save properly

So the problem is "badly behaved plugin can break bespoke locale fix, so do it when you stream"

KottV:
confirm, it fixes savestates broken by plugins that changes the locale

mvf:
The locale is global state, so I'm afraid this is subject to race conditions and won't reliably solve the problem. Better fix the code to call functions that don't rely on the locale, or pass in a locale explicitly.

baconpaul:
Sure! The best fix is to scrub the json library to not use the std lib to go to strings and write a locale independent formatter. 

But in the bespoke threading model this should improve the situation with badly behaving plugins 

locale being global is gross and causes problems; this shuffles a lot of them under the rug basicalky

baconpaul:
https://github.com/BespokeSynth/BespokeSynth/blob/b9d6dc394b623e099ede05fadd1753bd33a65619/libs/json/src/lib_json/json_writer.cpp#L70

We could I replace those lines with a std stream that we imbue wirh locale “C”

baconpaul:
Oh and the appropriate reader also of course

baconpaul:
https://github.com/BespokeSynth/BespokeSynth/blob/b9d6dc394b623e099ede05fadd1753bd33a65619/libs/json/src/lib_json/json_reader.cpp#L631 There’s the reader size that needs fixing 

good news is we have a clean repro now though

mvf:
> this shuffles a lot of them under the rug basically

Exactly, better fix it properly. Especially since you found that JUCE shuffles the locale on X11, we should try everything to be locale-independent.

I'd go with `std::to_chars`/`from_chars` as someone already suggested on Discord. It's locale independent and more efficient than `sprintf`. Let's just hope it's available on El Capitan.

baconpaul:
(Basically sigh libjson is not locale aware software :( so patch it or kludge the global state)

baconpaul:
#384 makes this unneeded

PR #374, opened 2021-10-10
Disable build-time version info by default, fix build when disabled

mvf:
I think you might care about this @baconpaul. This is a useful feature, but I don't think it's worth enabling by default for Bespoke.

---

The BESPOKE_RELIABLE_VERSION_INFO build option makes sure build info is
always up to date, but the flip side is that the build is always out of
date. This slows down debug sessions and can hide bugs in other CMake
rules that cause them to be always out of date unintentionally.

In a plugin (such as Surge, where this code came from), this tradeoff is
worth it because it's not always clear where the plugin gets loaded from
at any given time, and reliable build info by default can save lots of
grey hair.

But in a stand-alone application like Bespoke, it makes more sense to
default to a sane build environment that knows when stuff needs to be
rebuilt. If you're unsure what you're running, just look at the .EXE.


baconpaul:
That’s fine! I will just set it on in clion if I want it

baconpaul:
Oh if disabled does it show disabled or does it show a stale version? I think since people will self build if it is disabled we should show that and not a stale. this is mostly for the nightlies but self build does matter I guess!

baconpaul:
Oh and we should force it on in the pipelines of course - that’s the point of this :)

mvf:
The setting doesn't affect pipeline builds since they're always clean.

But I agree, it should be visible whether the info was actually taken at build or at configuration time. That wasn't the case before either though, because it didn't build with the setting off.

mvf:
Guess that's what happens when reading from the bottom up: In this PR's current form, pipeline builds have up-to-date data because they're clean, but on incremental builds the data will still be stale (just like before). If we change this to make it show the data only when taken at build time, then we'd force the option on for pipeline builds so they're dated again.

The point of this PR though is to fix the build and declutter the `CMakeLists`, not to reinvent the option (which I didn't submit).

baconpaul:
Ok cool that’s fine with me. Maybe we want to turn it so self build with disabled say disabled instead of potentially stale data? But that’s a separate and easy fix

PR #375, opened 2021-10-10
Minimap tweaks

NoxyNixie:
* Added bookmark buttons to the minimap.
* Fixed minimap positioning with different UI scales.
* Fixed module tooltips from showing up on modules without a title bar.
* Reduced the margin a little bit so it is more aesthetical with larger UI scales.
* Change Minimap to use unique_ptr.
* Made dragging on the minimap also work as opposed to clicking only.

I hope this all flies well but I know I'm not flawless and can take criticism or if you desire things to be different I'm up for changing things too ;)

### DIFF
--- a/Source/Minimap.cpp
+++ b/Source/Minimap.cpp
@@ -3,171 +3,265 @@
 #include "OpenFrameworksPort.h"
 
 const float MAX_LENGTH = 150;
-const float MARGIN_RIGHT = 20;
-const float MARGIN_BOTTOM = 20;
+const float MARGIN_RIGHT = 15;
+const float MARGIN_BOTTOM = 15;
+const float BOOKMARK_SIZE = 15;
+const float NUM_BOOKMARKS = 9;
 
-Minimap::Minimap() {}
+Minimap::Minimap()
+   : mClick(false)
+   , mGrid(nullptr)
+{}
 
 Minimap::~Minimap() {}
 
-bool Minimap::AlwaysOnTop() { return true; }
-
-bool Minimap::IsSingleton() const { return true; }
+void Minimap::CreateUIControls()
+{
+   IDrawableModule::CreateUIControls();
+   mGrid = new UIGrid(0, 0, MAX_LENGTH, BOOKMARK_SIZE, NUM_BOOKMARKS, 1, this);
+}
 
 void Minimap::GetDimensions(float& width, float& height)
 {
-    float windowWidth = ofGetWidth();
-    float windowHeight = ofGetHeight();
-    float ratio = windowWidth / windowHeight;
+   float windowWidth = ofGetWidth();
+   float windowHeight = ofGetHeight();
+   float ratio = windowWidth / windowHeight;
 
-    if (ofGetWidth() > ofGetHeight())
-    {
-        width = MAX_LENGTH;
-        height = MAX_LENGTH / ratio;
-    }
-    else
-    {
-        height = MAX_LENGTH;
-        width = MAX_LENGTH * ratio;
-    }
+   if (ofGetWidth() > ofGetHeight())
+   {
+      width = MAX_LENGTH;
+      height = (MAX_LENGTH / ratio) + BOOKMARK_SIZE;
+   }
+   else
+   {
+      height = MAX_LENGTH;
+      width = (MAX_LENGTH * ratio) + BOOKMARK_SIZE;
+   }
+}
+
+void Minimap::GetDimensionsMinimap(float& width, float& height){
+   GetDimensions(width, height);
+   if (width < height)
+   {
+      width -= BOOKMARK_SIZE;
+   }
+   else
+   {
+      height -= BOOKMARK_SIZE;
+   }
 }
 
 void Minimap::ComputeBoundingBox(ofRectangle& rect)
 {
-    std::vector<IDrawableModule*> modules;
-    TheSynth->GetRootContainer()->GetAllModules(modules);
+   std::vector<IDrawableModule*> modules;
+   TheSynth->GetRootContainer()->GetAllModules(modules);
 
-    if (modules.empty())
-    {
-        rect = TheSynth->GetDrawRect();
-        return;
-    }
+   if (modules.empty())
+   {
+      rect = TheSynth->GetDrawRect();
+      return;
+   }
 
-    rect = modules[0]->GetRect();
+   rect = modules[0]->GetRect();
 
-    for (int i = 1; i < modules.size(); ++i)
-    {
-        if (!modules[i]->IsShowing())
-        {
-            continue;
-        }
-        ofRectangle moduleRect = modules[i]->GetRect();
-        RectUnion(rect, moduleRect);
-    }
+   for (int i = 1; i < modules.size(); ++i)
+   {
+      if (!modules[i]->IsShowing())
+      {
+         continue;
+      }
+      ofRectangle moduleRect = modules[i]->GetRect();
+      RectUnion(rect, moduleRect);
+   }
 }
 
 ofRectangle Minimap::CoordsToMinimap(ofRectangle& boundingBox, ofRectangle& source)
 {
-    float width;
-    float height;
-    GetDimensions(width, height);
+   float width;
+   float height;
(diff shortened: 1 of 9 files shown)


awwbees:
I need to take the time for a deeper look at this, but two immediate notes:
- ModularSynth.cpp is missing a nullptr check on mMinimap, so you'll crash on load if the minimap is disabled 
- the bookmarks system seems pretty redundant with the LocationZoomer system, so I think it should integrate with that, or be removed if that's not feasible. the LocationZoomer is currently pretty invisible, so the minimap does seem like a pretty cool place to surface it

NoxyNixie:
I did see the LocationZoomer and in fact copied some of the code from there but after reading through it I tried its "keys" and nothing actually happened.

Now this might be a bug somewhere but failing to set "locations" I more or less assumed it was either working very awkwardly and I simply didn't know how, or it was broken or dead code (this assumption may be wrong).

The locationzoomer also has this slow lerping movement which to me for bookmarks is too slow, I wanted it to be instant.

I really like the minimap buttons and I'm fine with making them work with the locationzoomer data but I still want those things to be instant (And if we can make the hotkeys work and be instant too that would be neat).

But all that is probably for another day, heading to bed now ;)

awwbees:
yeah, I think it makes sense to modify the LocationZoomer to
offer the flexibility to achieve what you want, rather than having two systems that do nearly the same thing

awwbees:
it's also worth mentioning that I don't ever use LocationZoomer (and I doubt anybody does) so it is ok to change its functionality. although, I think the blending is good, because it makes it easier to follow the action. but the blending could be optional.

NoxyNixie:
I think I've identified why the hotkeys for map zooming aren't working. One of the main issues is with `GetKeyModifiers()`. It is quite confusing which keys are which when comparing mac and others.

I'm also not sure if juce is taking the right approach and mashing control into the `juce::ModifierKeys::isCommandDown()` on non mac.

I have a sneaking suspicion the whole `GetKeyModifiers()` and the Bespoke `KeyModifiers` enum are redundant and might only cause confusion since this enum does not match up with `juce::ModifierKeys::Flags`. For instance `juce::ModifierKeys::Flags::ctrlModifier` is always `2` while `juce::ModifierKeys::Flags::commandModifier` is either `8` or `2` depending on if its a mac or not. Now I could alter Bespoke's `KeyModifiers::kModifier_Command` to do something similar so that it reflects how juce does it but one might as well actually use the juce code directly instead.

If we at some point want to differentiate between left and right shift/alt/control/win/command keys we may have to not use juce since it appears juce simplifies all this to just control/alt/shift/command. Then again simplifying things is good too.

I've found two places where `KeyModifiers::kModifier_Control` was used one of which is most definitely a bug and likely the other one too so I've changed those to `KeyModifiers::kModifier_Command` instead but now the `KeyModifiers::kModifier_Control` value is completely unused. Should I change `KeyModifiers::kModifier_Command` to `KeyModifiers::kModifier_CommandOrControl` and set it whenever command or control (mac) or control (others) is set (And thus also removing `KeyModifiers::kModifier_Control` to avoid any further confusions)? This would no longer allow control and command to have different functions on mac but we already don't allow different functions on left and right key presses.

Ugh looking at the keyboard layout of a mac keyboard I would say `Alt == Command` but that is very different from what juce is doing. Or is alt actually a key on mac keyboards? I have no clue when it comes to mac's, never had one and never worked with one.

Either way my latest commit makes minimap bookmarks use the locationzoomer data and fixes the hotkeys ;)

awwbees:
sorry that I haven't gotten a chance to look at this yet.

here are the standard conventions for how mac -> PC modifier keys map:
- Command on a Mac keyboard->Control on a PC keyboard
- Option on a Mac keyboard->Alt on a PC keyboard
- Control on a Mac keyboard->most often, nothing

these are not just bespoke's conventions, they're pretty universally standard. for example, copy/paste on a mac is command-C/command-V

I'm generally not very comfortable with this PR for updating the minimap's behavior also touching the global key modifiers system... maybe it needs to be refactored eventually, but this PR isn't the right place for that.

NoxyNixie:
> * Command on a Mac keyboard->Control on a PC keyboard
> * Option on a Mac keyboard->Alt on a PC keyboard
> * Control on a Mac keyboard->most often, nothing

Ah that clear things up a little bit.

> I'm generally not very comfortable with this PR for updating the minimap's behavior also touching the global key modifiers system... maybe it needs to be refactored eventually, but this PR isn't the right place for that.

This PR only touches the minimap stuff right now. The whole keyboard convention questions were more for me to clear up my own confusion.

This PR changes none of the key behavior.

awwbees:
> This PR only touches the minimap stuff right now. The whole keyboard convention questions were more for me to clear up my own confusion.
> This PR changes none of the key behavior.

hmm, but I do see changes to SynthGlobals.cpp and LooperRecorder.cpp... were those accidental commits? I also see updates to the .gitignore which don't seem germane to this PR

NoxyNixie:
> hmm, but I do see changes to SynthGlobals.cpp and LooperRecorder.cpp... were those accidental commits? I also see updates to the .gitignore which don't seem germane to this PR

Whoops the .ignore was an accident. The SynthGlobals one is intended but might not pertain to this specific PR but it essentially the updated and as far as I can see the correct way to interface with JUCE. It saves on an allocation.

I can revert both or parts if you want and if you feel it nicer to deal with I could squash all the commits and make a "clean" commit message without all the steps in between?

awwbees:
ah! I understand now. yeah, if you could get rid of the gitignore (and also the LooperRecorder I think?) then I'll merge this. If you wanna squash it, that sounds good too.

NoxyNixie:
The LoopRecorder change is needed since its hotkeys (which seem to be global and not localized to that module) conflict with the LocationZoomer hotkeys. But that is only if you fix the seemingly incorrect usage of `kModifier_Control` in the LoopRecorder. If I revert that change the LoopRecorder hotkeys won't function on linux or windows and on mac will be the only thing that responds to hotkeys with the control key.

Source/Minimap.cpp
```diff
@@ -3,171 +3,265 @@
...
-    float width, height;
-    GetDimensions(width, height);
-    mX = ofGetWidth() - (width + MARGIN_RIGHT);
-    mY = ofGetHeight() - (height + MARGIN_BOTTOM);
+   float width, height, scale;
+   scale = 1 / TheSynth->GetUIScale();
+   GetDimensions(width, height);
+   mX = (ofGetWidth() * scale) - (width + MARGIN_RIGHT);
+   mY = (ofGetHeight() * scale) - (height + MARGIN_BOTTOM);
+   gDrawScale = gDrawScale;
```
awwbees: is this line a typo, or the weird result of a refactor?

Source/Minimap.cpp
```diff
@@ -3,171 +3,265 @@
...
+   UpdateGridValues(0, 0);
+}
+
+void Minimap::UpdateGridValues(float x, float y)
+{
+   mGrid->Clear();
+   for (int i = 0; i < mGrid->GetRows() * mGrid->GetCols(); ++i)
+   {
+      float val = 0.2;
+      if (TheSynth->GetLocationZoomer()->HasLocation(i + 49))
```
awwbees: what is the `49` here and on line 221?

awwbees:
I made a few of my own tweaks. I left a few questions above, and I may make some more tweaks.

Source/Minimap.cpp
NoxyNixie: It is the offset used by the LocationZoomer for storing its data. The locations are stored using their Hotkey char value.

Source/Minimap.cpp
NoxyNixie: That is wrong indeed and I don't remember why it is there :P

Source/Minimap.cpp
awwbees: switched it from the magic 49 number to '1'

awwbees:
alright! it's in. I moved it to the upper right corner to avoid colliding with the default position of the output modules, I guess I'll find out if that change bothers anybody

PR #376, opened 2021-10-10
Added Akai MPD232 midi controller mapping.

fcingolani:
Hello! I've added an MPD232 mapping. It should work with the "Generic" default presets in the MPD.

It includes:

- Stop, Play and Rec buttons support.
- 4 4x4 grids, one for every Pad Bank.
- 3 16 knob / 16 toggles / 16 slider groups, one for every Control Bank.

Captura de pantalla 2021-10-10 161609

Please let me know any questions or suggestions.

Thanks for creating Bespoke!


awwbees:
thanks!

PR #381, opened 2021-10-11
fix UI strangenesses when working while paused

awwbees:
- fix issue where script modules appeared to be constantly executing while paused
- fixed superfluous warnings about mono/stereo mismatches on cables


awwbees:
trying to figure out why all of these superfluous commits are ending up in here rather than just the final one... I am so dumb at git

PR #384, opened 2021-10-11
Move JSONCPP to 1.9.4; and submodule

baconpaul:
1. Move jsoncpp to a submodule inside libs/json but sill using our
   way stripepd back CMake above it in libs/json
2. Move to version 1.9.4
3. Upgrade the various things which broke around indexing
4. And confirm that this fixes the internationalization problem


baconpaul:
As discussed in discord lemme see if I can fix those deprecation warnin gs

baconpaul:
OK did a second push to fix all those deprecation warnings

awwbees:
looks good to me! did some testing and everything seems to function correctly.

I noticed one minor difference that lists seem to serialize with linebreaks now, so what once was
```
"position" : [ 44.0, 924.0 ],
```
is now
```
"position" : 
         [
            44.0,
            924.0
         ],
```

but, I can't think of any reason why this would be bad.

baconpaul:
Oh we can fix that with the StreamBuilder somehow. There's an option to turn it back. Is that branch the pretty branch of the non pretty branch?

awwbees:
that's on the pretty one

baconpaul:
darn it their (crummy) new api doesn't support control of compact rendering only the deprecated api does. That decision is now controlled on whether the total string length would be longer than `rightMargin_` which is a private variable with not accessor.

I really feel like they didn't make things better here and their deprecated API was superior to their new one :(

baconpaul:
Anyway alas there are no great options in front of us

1. merge my old change with a race condition
2. merge my original commit here with the old formatting and the new lib but deprecation warnings
3. merge this version with the new lib and new deprecation but different formatting
4. patch the library so we get the old formatting from the new code

ugh i hate all of those :(

I think 3 is the best but i understand if you prefer 2 or 4 or 1. or none of the above.

awwbees:
yeah, I think 3 is the best call. and then the better long-term call is to find a json library that is actually good.

baconpaul:
Yeah I thought about that but changing the base class of ofxJsonValue seemed like a lot of work! 

Ok I guess tbis is good to merge then unless you want other changes? Thanks!

PR #385, opened 2021-10-11
add mutex around access to the list of patch cable sources

awwbees:
fixes thread safety issues that seem to occur more readily in debug mode. this thread safety issue was potentially uncovered by #349


awwbees:
reopening as #393 to fix up the gross commit history on this one

PR #389, opened 2021-10-12
Add VS Code related extensions to .gitignore

ghost:
- VS Code config file extensions and folder
- CMake(Tools) default build folder


awwbees:
thanks!

PR #393, opened 2021-10-12
add mutex around access to the list of patch cable sources

awwbees:
fixes thread safety issues that seem to occur more readily in debug mode. this thread safety issue was potentially uncovered by #349


awwbees:
this doesn't fix the issue, there are other thread safety issues as well. I'm going to roll back #349 for 1.0.1, then take another stab later

PR #394, opened 2021-10-12
Fix Command Q on Mac with a Temporary Fix

baconpaul:
See the comment in MainComponent.cpp about the correct
fix but this gets 1.0.1 looking like 1.0 and restores
cmd q to quit on mac

### DIFF
--- a/Source/MainComponent.cpp
+++ b/Source/MainComponent.cpp
@@ -447,12 +447,11 @@ private:
    
    bool keyPressed(const KeyPress& key) override
    {
-#if BESPOKE_MAC
       /*
        * This is a temporary fix for 1.0.1. This keyPressed handler
        * always returns true whether or not Bespoke handles the event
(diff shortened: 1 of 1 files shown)


awwbees:
did this work on your machine? I had two issues with it:
- the keycode seemed to return 'Q' rather than 'q'
- the keyStateChanged() message was still swallowing the event, so a quit wouldn't occur

I pushed a change to fix these issues for me, and also add alt-f4 support, and tested on mac and windows. does it work on your machine with these changes?

baconpaul:
it did! Let me check your branch in a minute

baconpaul:
Oh damnit sorry the code i tested wasn't the code i pushed. I rewrote that comment and cleaned it up. Sorry bout that :( Sloppy.

You are correct your change is correct and my PR is wrong. But i confirm that it now works on my mac again.

The right thing of course is to make all the keypress handlers go through and return true false accordingly like i said. but this at least got it working.

PR #397, opened 2021-10-13
Fix Grid + Grid keyboard bugs.

NoxyNixie:
* Fix a crash when setting your scale to chromatic when you have a Grid keyboard module connected with a grid.
* Fixed Grid not actually accepting clicks on the notes in the scale when connected to a grid keyboard.

I'm not a 100% on if these are the correct fixes but from all my testing the results seem to be correct.

I don't have any actual grid controllers so unfortunately can't test those.

### DIFF
--- a/Source/Scale.cpp
+++ b/Source/Scale.cpp
@@ -133,7 +133,7 @@ void Scale::Init()
    
    Json::Value& scales = root["scales"];
 
-   mScales.push_back(ScaleInfo(kChromaticScale, std::vector<int>{}));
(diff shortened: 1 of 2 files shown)


awwbees:
sorry, no, neither of these are correct fixes:

1. "chromatic" mode intentionally gives a scale with zero elements, so it can dynamically handle non-12-pitch scales. hardcoding it to 12 pitches will break this. this was a recent change that happened in the scale overhaul to handle SCL/KBM loading. it looks like there are some cases that are not being handling the new style of chromatic mode correctly, so I'll fix those
2. this change will break grid functionality elsewhere. the actual issue here is that using the "grid" module as a click input is super broken in many cases, and I probably should have disabled it before release.

PR #399, opened 2021-10-13
Arturia Minilab MKII support

asmw:
In the out-of-the-box state my controller sends notes on channel 10 for the drum pads.
They can be toggled to send control messages via a button.

So I made two layouts which filter on the appropriate channels to make the whole thing usable


awwbees:
oh great! thanks for the channelfilter functionality

PR #400, opened 2021-10-13
work in progress to move VST scanning to another process

awwbees:
todo:
- [x] need to add CMakeLists stuff to get it to build properly
- [x] need to give it the path to launch for non-windows platforms
- [ ] need to evaluate if running the process via this method is the right way to go
- [ ] need to make the new VST scanner app not terrible, this is essentially just a stub

### DIFF
--- a/BespokeVSTScanner/CMakeLists.txt
+++ b/BespokeVSTScanner/CMakeLists.txt
@@ -1,0 +1,28 @@
+juce_add_gui_app(BespokeVSTScanner
+        )
+target_compile_definitions(BespokeVSTScanner PRIVATE
+        JUCE_DISPLAY_SPLASH_SCREEN=0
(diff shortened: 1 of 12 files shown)


mvf:
This change seems to add a new GUI app that does the plugin scanning. This approach is less than ideal:

- It splits the application into two executables (inflexible, maintenance burden)
- Starting a new application is messy and only needed on Windows. Elsewhere we can just `fork(2)`.
- The extra binary is very large (multi-megabyte). Reusing `Bespoke.exe` (e.g. via commandline parameter) avoids this bloat and is also easier to maintain.

I suggest first implementing the new UX, inside Bespoke, no separate app/process, and merge. Then the function that does the actual scanning can be quite easily moved to another process in a follow-up change.

awwbees:
I think #421 accomplishes these goals and obsoletes this PR

PR #405, opened 2021-10-14
Some light profiling on OutputChannel

baconpaul:
Did some light profiling. In a simple case on mac release build
of note sequencer -> osc -> output about 11% of the audio
time was in OutputChannel::Process -> Buffer::GetChannel.
Fix this by calling the get function outside a loop taking
the time to about 6%.


baconpaul:
Before:
Screen Shot 2021-10-14 at 2 19 07 PM

After:
Screen Shot 2021-10-14 at 2 27 27 PM

PR #408, opened 2021-10-15
Add increment epsilon for infinite encoders

asmw:
The infinite encoders on my Minilab send two messages for every move
one value in the 'direction', i.e. 0.5 +/- 0.1 and one 'centered' value
very close to, but not quite 0.5.

The absolute of the value must be larger than incrementEpsilon to affect
a change


awwbees:
could you rename "epsilon" to "threshold"? I feel like "epsilon" is too intimidating of a term to make user-facing.

asmw:
> could you rename "epsilon" to "threshold"? I feel like "epsilon" is too intimidating of a term to make user-facing.

Done.

PR #409, opened 2021-10-15
Add Akai MIDImix MIDI controller mapping

AlexMitchellMus:
Default controls mapped apart from `SEND ALL` which is not a MIDI control anyway. `SEND ALL` sends the current state of the whole device at once.

Screenshot from 2021-10-15 18-55-58


awwbees:
thanks!!

Issue #403, opened 2021-10-14
The "sustainpedal" module doesn't send the same note if it's already sustained.


dertasiu:
If I attach a "midicontroller" to a "sustainpedal" module and patch the 64 midi cc to the "sustain" checkbox on the "sustainpedal", I can't send the same sustained note twice unless I deactivate and reactivate the "sustain" checkbox.
For example, if the "sustain" checkbox is enabled and I press the C key twice, it will only sound the first time.

Tried on [07dd70b](https://github.com/BespokeSynth/BespokeSynth/commit/07dd70bb5925db271d59bd1805941fe86e13400e) commit, Linux with VST enabled and the Linux high FPS patch.

awwbees:
oh interesting, it looks like I did this intentionally, but I can't remember why:
`if (!mIsNoteBeingSustained[pitch]) //don't replay already-sustained notes`

if you comment out this line (line 73 of SustainPedal.cpp), does that give the behavior you expect?

dertasiu:
I didn't saw that line, my bad.
I've created a pull request (#411) with a checkbox to allow bypassing the current behavior.
Thank you!

awwbees:
fixed by #438

PR #411, opened 2021-10-15
Added an option to send already-sustained notes on sustainpedal

dertasiu:
In reference to the issue: #403 
I've added a checkbox to the sustainpedal module to allow sending already-sustained notes.
imagen

### DIFF
--- a/Source/SustainPedal.cpp
+++ b/Source/SustainPedal.cpp
@@ -28,7 +28,7 @@
 #include "ModularSynth.h"
 
 SustainPedal::SustainPedal()
-: mSustain(false)
+: mSustain(false), mNoteRepeat(false)
 {
 }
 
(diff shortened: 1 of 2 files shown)


awwbees:
I wonder if the option here is even necessary... maybe I should just remove my check and make note repeats always valid? can you think of any reason to retain the old behavior? it might have just been something that I didn't think hard enough about, and your desired behavior is more correct.

dertasiu:
Right now I can't find a use case for the old behavior. In fact, every software that I've used where the sustain pedal is supported has worked like the "new behavior".
When I saw that comment line, I thought that maybe you used the module like that, so I added the checkbox in case anyone was using it.
But now that you say that, I think you should remove it.

awwbees:
cool, that removal is now done in #438

PR #412, opened 2021-10-15
Change for Discussion nnot Merging

baconpaul:
So I don't think this is mergable - putting it here as a sort of
draft PR. Using the profiler on a heavy granular patch I found
more time in ChannelBuffer::GetChannel still. So what I did here
(which we can propagate if you agree) is add a GetChannelUnchecked
whcih is const inline direct access to the float with no if and no
alloc, and then used that. It took about 5% out of the CPU time
in the audio thread.

I think there's other performance hits in the granulator path also
and lots of little things like clamps and maps which don't inline
and stuff but pushing this up to see what you think of the idea.

Since I haven't checked every client of Grain::Process I don't
know which ones explicitly or implicitly guanrantee their buffers
so I probably wouldn't merge this

### DIFF
--- a/Source/ChannelBuffer.cpp
+++ b/Source/ChannelBuffer.cpp
@@ -86,6 +86,13 @@ float* ChannelBuffer::GetChannel(int channel)
    return ret;
 }
 
+void ChannelBuffer::GuaranteeChannel(int channel)
+{
+   // For now we can just call GetChannel which guarantees it
+   // but split the API just in case we want to do something else
+   GetChannel(channel);
+}
+
 void ChannelBuffer::Clear() const
 {
    for (int i=0; i<mNumChannels; ++i)
--- a/Source/ChannelBuffer.h
+++ b/Source/ChannelBuffer.h
@@ -37,6 +37,11 @@ public:
    ~ChannelBuffer();
    
    float* GetChannel(int channel);
+   void   GuaranteeChannel(int channel);
+   float* GetChannelUnchecked(int channel) const
+   {
+      return mBuffers[channel];
+   }
    
    void Clear() const;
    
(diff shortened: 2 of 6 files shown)


awwbees:
I think the `GuaranteeChannel()`/`GetChannelUnchecked()` interface feels weird (it seems like the sort of thing that I would stumble back across in 2 years and be like "what does this mean again?") but I like the idea, and what it does for performance. I'll think for a minute about an interface which pulls off the same thing but feels better to me.

the granulator optimizations definitely look good.

baconpaul:
yeah I didn’t quite know where to do it. I preferred having the initialization and access being separate but didn’t know where to put the init. But once you solve that you ca have a const in-line get channel which removes the branch and time and stack. 

 The unchecked is a big part of the granulstor thing because the process is per sample and calls getchannel so let me leave this to hang out while you ponder the right api and the. I can redo them. Sound good?

awwbees:
I have a change in https://github.com/awwbees/BespokeSynth/tree/preallocate-channels that might be a good way to transparently eliminate the branch in GetChannel(). I need to think harder about it to make sure it catches all cases, and do some testing, but this might be the way to go: preallocate ChannelBuffer buffers rather than trying to be clever about it and wasting time on the audio thread. the only time I want to avoid unnecessarily preallocating them is in the case of large Sample buffers, since that would cause significant waste, but that should all be handled by this change.

I ran it in the profiler and I got a good percentage knocked off the granulator process time, at least. did you have a good test case for profiling the GetChannel() change? would be nice to put it through its paces here.

baconpaul:
right that change looks like the pre-allocation i didn't know where to put has been put where it should go. Very good. That will solve many problems in an un-invasive way

https://www.dropbox.com/s/49uxb97rzblo5qo/Test_Ambient_sth.bsk?dl=0

there's the test patch I was using for granulator profiling which one of our discord users gave me. Your change should improve it quite a lot, and once its done, I can go re-apply my other changes (like the unbranched and faster window and stuff) if you see an improvement here.

Exciting!

baconpaul:
Closing this since 448 is merged and your channel fix is way better!

PR #413, opened 2021-10-15
Note Expression Router

baconpaul:
The Note Expression Router is a new module which is like
a note hocket but rather than having a propability per
bus, it has an expresison in two variables 'n' and 'v'
for the note and velocity, which allows you to choose
where to route the note.

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -183,6 +183,7 @@ target_sources(BespokeSynth PRIVATE
     NoteCreator.cpp
     NoteDelayer.cpp
     NoteDisplayer.cpp
+    NoteExpressionRouter.cpp
     NoteFilter.cpp
     NoteFlusher.cpp
     NoteGate.cpp
--- a/Source/ModuleFactory.cpp
+++ b/Source/ModuleFactory.cpp
@@ -247,6 +247,7 @@
 #include "MPETweaker.h"
 
 #include <juce_core/juce_core.h>
+#include "NoteExpressionRouter.h"
 
 #define REGISTER(class,name,type) Register(#name, &(class::Create), &(class::CanCreate), type, false, false);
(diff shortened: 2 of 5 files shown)


baconpaul:
Review greatly appreciated. I think I got everything right and it does indeed work as it says on the tin when I use it on my mac!

baconpaul:
Huh figuring out why scale doesn’t build on windows when I include exprtk is gonna be interesting - will report back tomorrow

baconpaul:
OK with that push this now builds for me. Lets see what CI does!

The thing I couldn't figure out is how to filter by channel, since it seems notes don't have channels? There's a voiceIdx - is the right idiom that I should make that available as an expression syntax then use a midi controller with channel as voice?I'm thinking how would I route a linnstrument in split mode, for instance, using this, and in `::PlayNote` I don't get a look at the midi channel of the note.

awwbees:
yeah, I don't pull channel data through with notes, which was a bit of an early oversight that I've never rectified.

on the midicontroller object, in the triangle menu, there is a "channelfilter", so you could have two midicontrollers to handle two different channels. that handles a lot of cases, but is unfortunately probably not appropriate for the linnstrument's multichannel MPE setup, you'd want a channel range filter instead, I'd think

baconpaul:
Right. So sounds like need something like INoteSource/INoteReciever get a `struct Note { p, v, c }` and then you add a virtual `PlayNote(const Pitch &) {}` which calls `PlayNote(n.p, n.v )` etc.... then upgrade the codebase slowly to be object vs argument aware. That sort of thing? Well that's more than we can do in this PR but I think this object is still useful for deterministic layering. If you agree happy for you to merge it?

And yes a channel range filter is exactly what I was thinking. Was hoping I could do it here with 'c' and I think it would be the right place, once 'c' is available to me :)

baconpaul:
Thanks for the improvements and the merge!

PR #424, opened 2021-10-18
allow .BSK files to be opened directly, or dropped onto the app window

### LINKED ISSUE #209: Associate .bsk files with bespoke on mac

awwbees:
from @Tadaboody's work on #216
fixes #209


Tadaboody:
Tysm! Sorry I never got to finishing it 😅

PR #426, opened 2021-10-18
Add a play/pause button to the transport module

asmw:
This allows for control of the global transport play/pause state
via a midi controller

The only issue I see is that the control activation indicator circle is frozen when the transport is paused
Screenshot from 2021-10-18 19-55-18


awwbees:
looks good to me! yeah, the circle freezes because all animations are driven by the global time, which pauses when the transport pauses. but that's a different issue to fix a different day!

PR #427, opened 2021-10-18
Fix chorddisplayer crash when using the chromatic scale

asmw:


asmw:
Crashes for me on current main branch with just a midicontroller -> chorddisplayer and scale set to chromatic when playing any chord

awwbees:
this fix is incorrect, it only works if you're using a 12-pitch tuning system. I use an empty list of pitches to indicate "chromatic", with a length that automatically matches the current tuning system. but, you're the second person to make a PR with this mistake, so I added a comment to explain this behavior. added a fix for this crash in #430

PR #437, opened 2021-10-19
Using a more precise "center" value for endless encoders

lucminah:
Using an Arturia Minilab MkII, turning an endless encoder to the left
slightly sends a MIDI CC value of 63, seen in Bespoke as 0.50, and
doesn't effect any change in the value it's mapped to control.
I imagine this is caused by .5f being a less precise "center" value
than is needed in this case. Using 64/127.0f instead, the issue is
fixed.

The incrementThreshold in the Minilab's layout also needs to be set to
0 so as to not block any value from having an effect.

### DIFF
--- a/Source/MidiController.cpp
+++ b/Source/MidiController.cpp
@@ -190,7 +190,6 @@ UIControlConnection* MidiController::AddControlConnection(MidiMessageType messag
    if (layoutControl != -1)
    {
       connection->mIncrementAmount = mLayoutControls[layoutControl].mIncremental ? 1 : 0;
-      connection->mIncrementThreshold = mLayoutControls[layoutControl].mIncrementThreshold;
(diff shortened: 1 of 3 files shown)


awwbees:
@lucasminah I made a change to simplify the code, and eliminate the increment threshold code. the real issue here is that this particular controller sends a "64" value from its infinite encoders for some reason, which none of my controllers do, so I wasn't handling it correctly. can you test my change with your controller and verify that it still works?

lucminah:
@awwbees That's curious, using a MIDI monitor it seems to send 63 and 65 for slight turns to the left and right respectively, but it does send a value of 64 with every move as well to denote the center value it's using I imagine.
Either way, just tested a build with your changes and yes, everything works. Thank you!

PR #455, opened 2021-10-22
make it possible to spawn LFOs directly

awwbees:
no longer only accessible via right clicking a slider


everythingability:
Is this what I came here for. I'm on a Mac and I can't control-click i.e right, to get an LFO to appear.

Issue #338, opened 2021-10-04
[Feature Request] Add project name to the end of the recording filename


spectromas:
It would be great if you could distinguish which project a recording came from. I like the current filename but perhaps the option to addend the project name at the end would be useful. I think an option for preappend or append would be great, both are handy but personally I prefer append if it has to be only one.

asmw:
I will take a look at his, sounds like a good idea

asmw:
Should be in the next version then :)

spectromas:
Thanks @asmw, that's amazing!

PR #459, opened 2021-10-22
Use a project file stem as the saved recording prefix

asmw:
Implements #338


asmw:
So MacOS has path::stem but not path::string? Weird

asmw:
Read the log wrong. Apparently no C++17 features on MacOS?

PR #460, opened 2021-10-22
Overflow grid x values in the drumplayer to subsequent rows

asmw:
This allows for better use of longer, single-row pad setups.

This allows the use of two drumsampler rows from an 8-pad, single line layout as found on the Arturia Minilab.

The alternative would be to tweak the layout which would then not reflect the physical device layout


awwbees:
looks good to me!

PR #461, opened 2021-10-22
Windows 32-bit: Restore support, fix crash on startup

mvf:
32-bit virtual address space is only 2GiB by default. Bespoke allocates
a ~330 MiB chunk on startup, which was failing with `bad_alloc`. So let
the linker set the `IMAGE_FILE_LARGE_ADDRESS_AWARE` flag, extending VA
space to up to 3 GiB on 32-bit systems (when booted with `/3G`) and the
full 4 GiB on 64-bit systems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,15 +25,6 @@ project(BespokeSynth VERSION 1.0.1 LANGUAGES C CXX ASM)
 message(STATUS "Build Type: ${CMAKE_BUILD_TYPE}")
 math(EXPR BESPOKE_BITNESS "${CMAKE_SIZEOF_VOID_P} * 8" OUTPUT_FORMAT DECIMAL)
 message(STATUS "Targeting ${BESPOKE_BITNESS} bit configuration")
-if (WIN32)
-    if (${BESPOKE_BITNESS} EQUAL 32)
-        message(WARNING "You are trying to do a 32 bit windows build, which is unsupported. This means"
-                " either your IDE tool chain is not set to 64 bits or your command prompt defaults to 32."
-                " It can usually be resolved by either editing your IDE settings to target a 64 bit build"
-                " or by removing this build dir and adding '-A x64' to your cmake command arguments")
-        message(FATAL_ERROR "32 bit build not proceeding")
-    endif()
-endif()
 
 # Global compile options
 add_compile_options(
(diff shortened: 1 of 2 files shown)


baconpaul:
I figured just explicitly don’t support 32 bit was a better strategy. Fine either way of course but is there a use case for a 32 bit daw that makes sense?

mvf:
> I figured just explicitly don’t support 32 bit was a better strategy.

The change that jettisoned 32-bit (on Windows only) didn't give any technical reason for doing so. 32-bit is no real extra effort, so I see no rush to drop it just yet. The use case for 32-bit Windows in particular is legacy VSTs. We also have Raspberry Pi users with 32-bit userland.

baconpaul:
64 bit windows will load 32 bit vsts. Like I said I’m not bothered. I just wanted to avoid people having bad self builds since 32 bit win daw just isn’t useful won’t scan most vsts may not link and load python etc… so try and make the build you probably don’t want show an error was my thinking.

mvf:
> 64 bit windows will load 32 bit vsts.

Yes, but only into a 32-bit process. The big DAWs work around this by shipping two plugin hosts.

Of course I'm well aware that 32-bit is almost dead and wouldn't suggest shipping 32-bit releases. But unless it causes extra effort there's no harm in letting tinkerers decide for themselves what's useful. Build issues, e.g. with Python, can just as easily occur when building for 64-bit, but IME the experience is basically perfect. I have both a 32-bit and a 64-bit Python installed, and CMake reliably detects the right one.

baconpaul:
Ha yeah sure. So my problem was I didn't have 32 bit python installed and was doing an accidental 32 bit build so added a warning. Then figured why not just make it fatal. 

At least now we show the bittiness at build time so I would have caught it. 

And agree 32 bit windows should die. I'm thinking about no longer shipping a 32 bit surge binary with our next release since our downloads are just so few.

mvf:
Gotcha. Yeah, the bitness CMake message brought some much-needed clarity, also very useful in cross builds. Totally agree that shipping 32-bit Surge XT is not worth the effort, all the more since we're not a plugin host.

Gotta say though, when debugging this 32-bit Windows stuff, the short pointers were really pleasant. I might switch back to 32-bit for my Windows work. :sweat_smile:

PR #466, opened 2021-10-23
Picking up patchcables and right clicking disconnects them

asmw:
Makes cable management with the mouse much more fluid for me


awwbees:
I have a different take on this in #468 which I think is a bit cleaner and functions better (it does the delete on right mouse down instead of right mouse release), do you want to check if that one works for you?

asmw:
#468 is the way to go

PR #467, opened 2021-10-23
Fix control cable disconnects via backspace

asmw:
This will conflict with my 'right-click-to-disconnect' changes.

If both changes are acceptable, I will take care of that

### DIFF
--- a/Source/MidiController.cpp
+++ b/Source/MidiController.cpp
@@ -1331,6 +1331,18 @@ std::string MidiController::GetDefaultTooltip(MidiMessageType type, int control)
    return str + ofToString(control);
 }
 
+UIControlConnection *MidiController::GetConnectionForCableSource(const PatchCableSource *source)
+{
+   for (auto c : mLayoutControls)
+   {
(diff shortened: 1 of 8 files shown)


awwbees:
can you further describe which issues this fixes? I want to be able to test and see.

asmw:
> can you further describe which issues this fixes? I want to be able to test and see.

Ah sorry, I thought this was a known issue.

I cannot disconnect connections from a midicontroller button by picking it up and pressing backspace.
image

The ONLY way I can remove a midicontroller connection is by clicking the 'x' in the list or in the individual connection panel.
image

awwbees:
I'm gonna have to come back to this and do a little more work before merging. with this functionality, it puts the midicontroller into a weird state where the cable is disconnected, but the control mapping entry is still hanging around. I think the correct fix will be to make deleting the cable fully equivalent to hitting the "x" button

asmw:
Ah right, didn't see that, sorry

awwbees:
looks good! sorry that it took me so long to merge this one, github's diff was making the change look much more significant than it actually was.

PR #468, opened 2021-10-23
delete held cable if you right-click

awwbees:
a different take on #466
one issue that I dislike about this is that you can't right click while dragging with the left mouse button down. there seems to be something deeper within juce's mouse handling that prevents additional button clicks from getting through when one mouse button is already held down. needs further research, I think it would be cool to be able to right-click-delete while the left mouse button is held!


asmw:
Works fine for me, I'll delete my PR

PR #476, opened 2021-10-24
Fix missing first notes on notecanvas when resetting the trasnport

### LINKED ISSUE #361: Resetting Timeline to Zero, first note is not played
Resetting timeline to zero through the transport module causes all notes set to time 0 to not be played.

To reproduce:
1. Create note canvas, attached to any synth and output.
2. Assign a note to the very start of the canvas.
3. Press "reset" on the transport module

A note should play, but it does not. The note does play as the canvas loops around, but not when the timeline is reset.

asmw:
I could produce the issue by recording from a midi device, quantizing
and moving the notes to the front of the canvas

With the smaller rewind the first notes would not play.
When placing notes with the mouse, the issue did not happen

Maybe fixes #361

### DIFF
--- a/Source/Transport.h
+++ b/Source/Transport.h
@@ -121,7 +121,7 @@ public:
    void SetMeasure(int count) { mMeasureTime = mMeasureTime - (int)mMeasureTime + count; }
    void SetDownbeat() { mMeasureTime = mMeasureTime - (int)mMeasureTime - .001; }
    static int CountInStandardMeasure(NoteInterval interval);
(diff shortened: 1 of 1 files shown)


asmw:
Not sure if doing it by nudging floats is reliable though

awwbees:
yeah, it's probably not the best solution and needs something better (the correct value is probably the length of an audio buffer in measure-time), but for the short-term, this helps and doesn't hurt anything else

PR #477, opened 2021-10-24
Add `BESPOKE_PORTABLE` build option

mvf:
When `BESPOKE_PORTABLE` is enabled, the build installs Python to a directory `python` in the output directory (next to `resources`), or into the bundle on macOS. This is somewhat unorthodox. Usually dependencies are bundled at install time. I went with this approach for now, for two reasons:
- It's dead-simple. You build, the output directory has everything, the end.
- It enables debugging portable builds directly from the IDE.

In portable builds, `ScriptModule` uses only the bundled Python and does not display a "make sure you have Python installed" prompt. All platforms also ship a standalone Python interpreter that targets the bundled Python. `pip` is supported via `ensurepip`. Just add a Script module and run:
```python
import ensurepip
ensurepip._main()
```

Platform notes:

**Windows**: Python is deployed as a side-by-side assembly. If the DLL is missing, Windows displays an error message and refuses to start Bespoke. There is no fallback to whatever the user might have installed. Tested with 3.9 (64-bit), 3.10 (32-bit).

**macOS**: The existing `nightly-package` target now produces a `.dmg` with a `Bespoke.app` that works on python-less systems. The uncompressed bundle is currently a tad heavy (~170 MB) with some Python versions. We can probably exclude more stuff during the install step. Tested with Apple 2.7, CommandlineTools 3.8, python.org 3.10.

**UNIX/Linux**: Python is referenced via relative rpath, so there's a fallback to the system python. Won't help much though since `ScriptModule` will insist on the portable `PYTHONHOME`. Tested with 3.10.

The next step is to generate installers for Windows and Linux.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,7 @@ cmake_policy(SET CMP0091 NEW)
 
 # Global build parameters. Optional features are configurable via additional
 # BESPOKE_* variables defined in their respective subdirs.
+option(BESPOKE_PORTABLE "Create a self-contained build that ships all dependencies" OFF)
 set(BESPOKE_JUCE_LOCATION "${CMAKE_SOURCE_DIR}/libs/JUCE" CACHE STRING "Path to JUCE library source tree")
 set(BESPOKE_PYTHON_ROOT "" CACHE STRING "Python search path. Override as needed.")
 set(BESPOKE_VST2_SDK_LOCATION "" CACHE STRING "Steinberg VST2 SDK directory for non-FOSS builds")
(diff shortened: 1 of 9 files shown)


mvf:
I can't reproduce the macOS build failure locally. It seems the `Python.framework` on the pipelines is wonky and contains unsigned binaries.

awwbees:
any risk here with merging this prior to the upcoming release? it looks like there are zero code changes when that define isn't set, but I'm not familiar with any of the other changes to be able to understand the impact

mvf:
There is no impact with the option off. Hopefully we can use it for the release though, and put an end to all those Python crash reports. My local build can already generate a (crude) WiX `.msi` from within CMake.

PR #478, opened 2021-10-24
starting a pass at cleaning up userprefs system

awwbees:
doing a test to make sure this compiles on all platforms, not just MSVC


awwbees:
cool, gonna keep working on it now

PR #481, opened 2021-10-25
Map all played notes onto a drum pad

asmw:
For pads which send notes which lie beyond the first 16 notes

I did not see a reason for some notes not to make a sound

### DIFF
--- a/Source/DrumPlayer.cpp
+++ b/Source/DrumPlayer.cpp
@@ -507,31 +507,29 @@ void DrumPlayer::PlayNote(double time, int pitch, int velocity, int voiceIdx, Mo
       return;
    }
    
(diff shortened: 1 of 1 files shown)


asmw:
The diff looks weird, I just replaced the modulo and removed the range guard. Everything else should be the same

awwbees:
nah, I did it this way because a lot of drum midi controllers tend to work in octaves, so it made more sense to modulo by a multiple of 12

asmw:
Would it be OK to do (pitch % 24) % NUM_DRUM_HITS?

This makes all of the 8 pads on my keyboard work using a midi connection without having to rewrite some of the note outputs

PR #497, opened 2021-10-28
when there is a hovered UI control, force the mouse to interact with it

### LINKED ISSUE #151: Highlights don't always match click targets

awwbees:
fixes #151


awwbees:
need to test this one a bit more, if there are any bugs they would be pretty bad, since it could block click interactions

awwbees:
seems pretty good so far. let's merge it and see if anyone finds a bug!

PR #505, opened 2021-10-30
linux fps: undo #396 and restore #349 for linux only

awwbees:
after testing, I and others can't reproduce the thread safety crash of #396 on linux, only on mac and windows. something must be going on under the juce hood with threading, to make this setting work for linux with my thread model, and bad for mac/windows.

long-term the real issue is safety issues in my thread model, so that should be fixed. but this seems to be safe in the short-term.


awwbees:
alright, screw it: I'm not 100% sure that this is safe, but I'm going to merge this for the 1.1.0 beta and see if any linux folks get crashes

PR #515, opened 2021-10-31
Fix notecanvas connections not disconnecting

### LINKED ISSUE #238: Backspace to unlink doesn't work on note canvas
As ridiculous as it sounds, yes.

I unlink other modules by grabbing the handle and pressing backspace. They (the couple ones I tested at least) work. But when I try to unlink a note canvas from an oscillator, it just... doesn't work.

Version: 1.0.0
OS: Arch Linux

(IMO this shortcut is quite inconvenient. I'd suggest something like Blender's ctrl-drag for cutting through conenctions. In general, there's something to be learned from blender's controls for nodes, but this is just my 2c).

Thank you :v:

asmw:
IDrawableModule::KeyPressed needs to be called for the backspace
on the cable to work.
The canvas will still get the keypresses this way.
The `GetLastClickedModule` condition was removed, because
drag-selecting notes in the canvas will not set that properly.
So just placing the mouse over a canvas will allow the key interaction.

This should fix #238

### DIFF
--- a/Source/Canvas.cpp
+++ b/Source/Canvas.cpp
@@ -542,7 +542,7 @@ bool Canvas::MouseScrolled(int x, int y, float scrollX, float scrollY)
 
 void Canvas::KeyPressed(int key, bool isRepeat)
 {
-   if (TheSynth->GetLastClickedModule() == GetParent() && gHoveredUIControl == this)
+   if (gHoveredUIControl == this)
    {
       if (key == juce::KeyPress::backspaceKey || key == juce::KeyPress::deleteKey)
       {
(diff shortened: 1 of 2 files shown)


awwbees:
fixing this with #521 instead

asmw:
@awwbees #521 is not a correct fix. Now the canvas widget gets the keypress twice if the module is `LastClicked` and the canvas is hovered, which makes the notes move twice the amount in that case

To reproduce:
- Make a notecanvas and add a note
- Make sure it is lastclicked and place the mousecursor on the canvas widget
- Press an arrow key (shift-arrow is even more noticable)

I think the whole `hovered` logic is confusing here

PR #533, opened 2021-11-02
Make the EnvelopeEditor behave like the LFO w.r.t. delete

### LINKED ISSUE #491: Pinned envelope editors can't be recreated
Using the build from 1a1593f0:

1. Create a patch that can utilize an `envelopeeditor`, like `oscillator`
2. Create an `envelopeeditor`, like on the corner of the `volume` box
3. Press the button to `pin` it
4. Select the `envelopeeditor` and then remove it by pressing backspace
5. After this, the `envelopeeditor` can't be recreated using the above steps unless you delete and recreate the patch

asmw:
Fixes #491 

... but I don't understand how these classes will ever be deleted? Who is responsible? The module factory?

### DIFF
--- a/Source/EnvelopeEditor.cpp
+++ b/Source/EnvelopeEditor.cpp
@@ -441,6 +441,12 @@ void EnvelopeEditor::SetADSRDisplay(ADSRDisplay* adsrDisplay)
    mMaxSustainSlider->SetShowing(mADSRDisplay->GetADSR()->GetHasSustainStage());
 }
 
+void EnvelopeEditor::DoSpecialDelete()
+{
+   mEnabled = false;
+   mPinned = false;
+}
+
 void EnvelopeEditor::DrawModule()
 {
    if (!mPinned)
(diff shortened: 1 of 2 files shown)


asmw:
Huh, just found this:
```
   //delete module; TODO(Ryan) deleting is hard... need to clear out everything with a reference to this, or switch to smart pointers
```

awwbees:
> ... but I don't understand how these classes will ever be deleted? Who is responsible? The module factory?
> 
> Huh, just found this:
> 
> ```
>    //delete module; TODO(Ryan) deleting is hard... need to clear out everything with a reference to this, or switch to smart pointers
> ```

yeah, I think as you found in your later pull request, modules are intended to be deleted when you reset the layout or load a new patch. truly deleting a module while a patch is running is just asking for a crash, since there are lots of references to shared memory that would need to be sanitized. I prefer to waste memory than risk the crashes and add headache to my life!

PR #534, opened 2021-11-02
Fix deletion of modules at least on patch reset

asmw:
In all places where `OnModuleDeleted` is called the module is marked as deleted directly before calling, so either the condition is wrong, or the code can be deleted


awwbees:
that check is to avoid double-deletions, and is correct. you are correct that there is a logic error that was causing deletions to be ignored, though, due to when I was calling `MarkAsDeleted()`. #542 is the correct fix, I will probably need to test it a bit before merging.

PR #536, opened 2021-11-02
Transport priority

ghost:
I've added a value to the interfaces which look at the transport time. This solves a couple of precise timing based issues i've come across, and allows for easy control of a component's execution order in the future.

to help with this, I've moved the time listeners to occur after the poll listeners. This fixes an issue where the event canvas would skip the first note of a sequencer when used to enable/disable them.

I've made a couple small changes to the pulse train as well, which will allow it to send a pulse received from other pulses tied to a time event, without immediately triggering a second pulse from its own time event.
I also made a change which will make the pulse train highlighted column rest on the first box, instead of after the last, because this seems like a better visual experience to me?

### DIFF
--- a/Source/EventCanvas.cpp
+++ b/Source/EventCanvas.cpp
@@ -50,7 +50,7 @@ EventCanvas::EventCanvas()
 , mPreviousPosition(0)
 {
    SetEnabled(true);
-   setPollPriority(0.1);
+   setPollPriority(0.3);
    
    mRowColors.push_back(ofColor::red);
    mRowColors.push_back(ofColor::green);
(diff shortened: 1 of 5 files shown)


awwbees:
I like this idea! it's maybe a liiitle hacky, but I think it's a good kind of hack. I should probably replace it someday with some sort of dependency-tracing system to automatically figure out update order.

I will hold off on merging this before my 1.1.0 release, because I think there's nonzero chance that reordering the `UpdateListeners()` and `OnTransportAdvanced()` could break an assumption and have some undesired impact, and I'll want time to test for that.

ghost:
yeah i have the exact same concern about reordering those two, so i'll just be testing it passively as i use bespoke going forward.

awwbees:
finally solved this problem in a different way in #972

PR #542, opened 2021-11-03
fix issue where modules weren't deleting

awwbees:
I was calling `MarkAsDeleted()` before `OnModuleDeleted()` (must have been a refactor gone awry), which was resulting in the deletion code being ignored, because it was tripping the check to avoid double-deletions


awwbees:
needs some testing before merging

PR #549, opened 2021-11-04
Portable Changes

baconpaul:
1. azure-pipelines builds portable
2. portable pips in jedi

please don't merge this - i'm using pipelines to test quickly


baconpaul:
OK @mvf would love your feedback on this. I've tested it on mac and win and it gives us a portable install with jedi and mido included (I picked mido pretty arbitrarily).

I'm rebuilding my linux VM (sigh) right now and then will test this on linux too but the pipeline worked.

@awwbees which packages do you think we want to pip in?

Also do you think we should run pip upgrade after we do enable pip or are the warnings OK?

baconpaul:
Sigh
this doesn't work in linux
```
ensurepip is disabled in Debian/Ubuntu for the system python.

Python modules for the system python are usually handled by dpkg and apt-get.

    apt install python3-<module name>

Install the python3-pip package to use pip itself.  Using pip together
with the system python might have unexpected results for any system installed
module, so use it on your own risk, or make sure to only use it in virtual
environments.
```

but it does work win/mac

so how to pip into the portable install on linux?

awwbees:
> @awwbees which packages do you think we want to pip in?

probably just jedi, I think

baconpaul:
Ok! Well I don’t quite know how to make ubuntu do the pip - if I did I could clean up this pr to mergeable. Any ideas? Or I could just pip Mac windows for now?

awwbees:
linux is so annoying. so is there no way to pip in this portable setup on linux? or is there at least a "user has to jump through additional hoops" way?

baconpaul:
> linux is so annoying. so is there no way to pip in this portable setup on linux? or is there at least a "user has to jump through additional hoops" way?

I don’t know those hoops. I think this is a choice ubuntu makes with their system Python but I’m not sure

@mvf any ideas? If not I can hack a bit more tomorrow

baconpaul:
OK played around some more

pip3 is a script on linux. so if i copy that script and run it with the local bespoke interpreter it runs, but it puts the resulting pipped content in ~/.local/lib/python3.8 not in the bundle

it seems there is no site-packages in the bundle on linux

but it also means I can't figure out how to make an in-package pipper

i wonder if we should just not do portable in linux and only do it on mac/win? then tell people if they use the binary they have to self build?

Source/cmake/bundle-python.cmake
```diff
@@ -41,15 +43,23 @@ if(APPLE)
     relinkPySigned("${pyExecutable}" "@executable_path/../${pySOName}")
     relinkPySigned("${pyDirDst}/Resources/Python.app/Contents/MacOS/Python" "@executable_path/../../../../${pySOName}")
     execute_process(COMMAND codesign -s - --deep "${bundleContents}/Frameworks/${pyFWName}")
+    execute_process(COMMAND "${pyExecutable}" -m ensurepip  --default-pip)
+    execute_process(COMMAND "${pyExecutable}" -m pip install ${BESPOKE_PIP_PACKAGES})
```
mvf:
```suggestion
```

Source/cmake/bundle-python.cmake
```diff
@@ -41,15 +43,23 @@ if(APPLE)
...
 elseif(WIN32)
     get_filename_component(pyDirSrc "${Python_EXECUTABLE}" DIRECTORY)
     file(GLOB pyBinaries LIST_DIRECTORIES false "${pyDirSrc}/*.dll" "${pyDirSrc}/*.exe")
     file(INSTALL ${pyBinaries} DESTINATION "${pyDirDst}")
     file(INSTALL "${pyDirSrc}/DLLs" DESTINATION "${pyDirDst}")
     file(INSTALL "${CMAKE_CURRENT_BINARY_DIR}/python.manifest" DESTINATION "${pyDirDst}")
+    set(pyExecutable "${pyDirDst}/python.exe")
+    execute_process(COMMAND "${pyExecutable}" -m ensurepip  --default-pip)
+    execute_process(COMMAND "${pyExecutable}" -m pip install ${BESPOKE_PIP_PACKAGES})
```
mvf:
```suggestion
```

Source/cmake/bundle-python.cmake
```diff
@@ -41,15 +43,23 @@ if(APPLE)
...
 else()
     get_filename_component(pySO "${Python_LIBRARIES}" REALPATH)
     file(INSTALL "${pySO}" DESTINATION "${pyDirDst}/bin")
     execute_process(COMMAND ${CMAKE_COMMAND} -E copy "${Python_EXECUTABLE}" "${pyDirDst}/bin/python")
     execute_process(COMMAND patchelf --set-rpath "$ORIGIN" "${pyDirDst}/bin/python")
+    set(pyExecutable "${pyDirDst}/bin/python")
+    execute_process(COMMAND "${pyExecutable}" -m ensurepip  --default-pip)
+    execute_process(COMMAND "${pyExecutable}" -m pip install ${BESPOKE_PIP_PACKAGES})
 endif()
```
mvf:
```suggestion
endif()

unset(ENV{PYTHONHOME})
execute_process(COMMAND "${pyExecutable}" -m ensurepip  --default-pip)
execute_process(COMMAND "${pyExecutable}" -m pip install ${BESPOKE_PIP_PACKAGES})
```

mvf (changes requested):
Thanks for looking into this @baconpaul. Sorry for my absence lately, things are slightly crazy over here. :hurtrealbad:

This change works for me on Linux. `ensurepip` generates the `pip` scripts in `<exedir>/python/bin` and `jedi` gets installed into the portable `python` as expected.

Only two things I can imagine breaking things:
- `patchelf` not being installed so `python` links against (and targets) the user's Python.
- `PYTHONHOME` being set in the build environment. `BespokeSynth` ignores `PYTHONHOME` since it explicitly calls `Py_SetPythonHome`, but `python` honors it. To guard against this, we should just clear it via `unset(ENV{PYTHONHOME})`, or alternatively, set it to the bundled Python. Clearing should be fine though, because if our Python doesn't work with `PYTHONHOME` unset, it won't work on user machines either. Maybe we can even build a sanity check around this: Install packages, then check if they are in our `site-packages`.

So this approach looks fine. Just suggested adding the `unset` and move the duplicated `execute_process` calls after the `if`. Didn't have time to test further though.

baconpaul:
ahh the unset is clever.

ignore that force push i just did.

baconpaul:
So @mvf the ensurepip thing is nothing to do with python home. it's that the ubuntu/debian python3 purposefully disables ensurepip to avoid you pipping into the system python and forcing you to venvs. Whether this is a smart decision or not i'm not sure, but it means if we build on ubuntu and copy their python we get the pip stage failing, as you can see in the azure log here, even with the PYTHON_HOME unset (or set to local location which i tested in my vm).

Basically: if you build portable this way on ubuntu you get a non-relocated-pippable python since if you install python on ubuntu it is non-pippable with ensure-pip. If you run the pip3 script with the embedded python it installs into your ~/.local directory not into the bundle.

I'm not necessarily sure this is a problem, but it does mean that portable python linux built on ubuntu doesn't work if you want to pip into it as opposed to your system, so a person with p39 who uses a portable bespoke with p38 could get library conflicts in their python if they pip.

I'm sort of leaning towards the approach of 'only do portable binary builds on mac and windows' and then use @KottV 's app images built non-portably and ability to self build on linux, since there's too many configurations to support and our build hosts here are ubuntu with this restriction. And since some (but not all) distros work with this pip change, merge it but don't use it for the official binaries (or maybe don't even have official binaries on linux).

Which sort of makes me think the right course of action is

1. merge this
2. awwbees does a win/mac binary build with portable on and lets people have a beta. we tell the linux community to do a build at head or use the app images and test
3. get test feedback and adjust the code
4. do the 1.1 non beta release based on 3, probably using the same strategy as 2, but perhaps also making a portable binary on linux if we can resolve this ubuntu pip restriction

thoughts?

baconpaul:
Right so in `/usr/lib/python3.8/ensurepip/__init__.py` on ubuntu

```python 
def _ensurepip_is_disabled_in_debian_for_system():
    # Detect if ensurepip is being executed inside of a python-virtualenv
    # environment and return early if so.
    if hasattr(sys, 'real_prefix'):
        return

    # Detect if ensurepip is being executed inside of a stdlib venv
    # environment and return early if so.
    if sys.prefix != getattr(sys, "base_prefix", sys.prefix):
        return

    # If we've gotten here, then we are running inside of the system Python
    # and we don't want to use ensurepip to install into the system Python
    # so instead we'll redirect the user to using dpkg and apt-get.
    print('''\
ensurepip is disabled in Debian/Ubuntu for the system python.

Python modules for the system python are usually handled by dpkg and apt-get.

    apt install python3-<module name>

Install the python3-pip package to use pip itself.  Using pip together
with the system python might have unexpected results for any system installed
module, so use it on your own risk, or make sure to only use it in virtual
environments.
''')
    sys.exit(1)

```

PR #560, opened 2021-11-05
Force win MSVC runtime to static linkage

baconpaul:
by setting MSVC_RUNTIME_LIBRARY, resolving the lack-of-msvc-dll
reported on discord this morning

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,8 @@ add_compile_options(
     $<$<CXX_COMPILER_ID:MSVC>:/utf-8>
     )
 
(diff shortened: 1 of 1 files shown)


baconpaul:
Confirmed this works; as @mvf says with portable we may not need it since python will ship the dll also, but I think it is a smart choice and one most audio software makes (and one we would need when we make bespoke a plugin almost definitely)

baconpaul:
hold off for a minute - @mvf had a good suggestion

baconpaul:
OK @awwbees this one is mergeable if you agree. Think we should merge it pre-beta

PR #581, opened 2021-11-07
Restore the non-portable python code [WIP do not merge]

baconpaul:
since signing the portable eludes us


baconpaul:
We won’t need this now we have signing working (phew!)

PR #582, opened 2021-11-07
Working signature with hardened runtime macos

baconpaul:
This makes it so you can build a signed macos portable but
you do it slightly differently. You build with signing of
then run the custom sign commands at the end of the build phase.

A typical session rather than setting CMAKE_XCODE_ATTRIBUTE_SIGN_AS would be

cmake -Bignore/macBLAH -DBESPOKE_PYTHON_ROOT=/Library/Frameworks/Python.framework/Versions/3.9/ -DBESPOKE_PORTABLE=True -DBESPOKE_SIGN_AS="Apple Development: you@yourspot.com (AFDFS23123)"
cmake --build ignore/macBLAH --parallel 8 --target BespokeSynthSigned


awwbees:
this worked and resulted in a notarized build and it's amazing

PR #585, opened 2021-11-08
Midi Controller sends CC out the MIDI pipe

baconpaul:
Send idi CC controllers out the midi output from MidiController


baconpaul:
@awwbees I don't know if you want this, but per discussion on discord, if you do, here it is!

Issue #590, opened 2021-11-09
Offer possibilities to devendor used libraries and tooling


dvzrv:
When building the package for Arch Linux I currently have to devendor some libraries by patching various CMakeLists.txt files (see relevant patch https://aur.archlinux.org/cgit/aur.git/tree/bespokesynth-1.0.0.r246.gfd389e6-devendoring.patch?h=bespokesynth-git).

Many libraries and tooling (e.g. jsoncpp, JUCE, pybind11) are available on downstream distributions and they usually prefer to build against those, as vendoring is bad (Fedora has an extensive write-up on the topic: [Bundled Libraries](https://fedoraproject.org/wiki/Bundled_Libraries)).

It would be awesome to be able to provide options for the use of system libraries (e.g. `USE_SYSTEM_JSONCPP`), which can be provided during build time to allow downstream distributions to seamlessly build against their system libraries instead of the bundled ones.

awwbees:
I don't really understand any of this (or am particularly interested in doing the research!) but I'm happy to take a pull request

dvzrv:
Cool! I will try to offer this as time permits!

NotAFile:
I would advise introducing some easy way to tell official builds from distro builds so that issues caused by subtle library version mismatches resulting from this can be correctly reported to the distros instead of here.

dvzrv:
Unfortunately there is not really a "one size fits all" solution to that problem.

One way to deal with it in the context of this project is to use [issue templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository), so that people have to report
* what version of the software they are using
* on what operating system (and optionally version) they are using the software
* how they installed the version that they are using

That way it becomes much clearer for the project developers to tell things apart and to e.g. ask for help if there is an issue with a specific distribution package.
It is important to not only look at this in the context of dependency issues, but also in the context of changes to the build system (which may introduce problems as well).

Users of the Arch Linux distribution are expected to follow the distribution's [bug reporting guidelines](https://wiki.archlinux.org/title/Bug_reporting_guidelines). [This section](https://wiki.archlinux.org/title/Bug_reporting_guidelines#Upstream_or_Arch?) in particular is relevant for this discussion. Having guidelines of course does not keep anyone from disregarding them and filing tickets in the wrong place, but there is not much that can be done about that I am afraid.

dvzrv:
[...]

PR #637, opened 2021-11-21
Allow devendoring of pybind11

dvzrv:
CMakeLists.txt:
Add the option `BESPOKE_SYSTEM_PYBIND11` to indicate that a system
installed pybind11 should be used instead of the bundled version.

libs/CMakeLists.txt:
Depending on `BESPOKE_SYSTEM_PYBIND11` use the pkgconfig integration to
find a system installed pybind11 or use the bundled pybind11.

Relates to #590


baconpaul:
Awwbees same comment on this one. Looks fine and our builds still use bundled.

PR #641, opened 2021-11-23
Allow devendoring of tuning-library

dvzrv:
CMakeLists.txt:
Add the option `BESPOKE_SYSTEM_TUNING_LIBRARY` (defaults to `OFF`) to
indicate that a system installed tuning-library should be used instead
of the bundled version.

Source/CMakeLists.txt:
Use cmake conditional generator expression to only link against the
bundled tuning-library if `BESPOKE_SYSTEM_TUNING_LIBRARY=OFF`.

libs/CMakeLists.txt:
Add conditional statement to only include bundled tuning-library if
`BESPOKE_SYSTEM_TUNING_LIBRARY=OFF` and otherwise attempt to find the
required header files in a system context.

Relates to #590


dvzrv:
If any of this needs rebasing or changes, let me know!

dvzrv:
@awwbees can you let me know, if the change is okay like this? I would like to include it in a package for Arch Linux, so that we can have it in the official repositories.

awwbees:
looks good!

PR #650, opened 2021-11-25
Fix for building with clang and clang-cl on Windows

mthierman:
This fixes building with clang and clang-cl compilers in Windows.


awwbees:
thanks! sorry for the very slow turnaround on this... got busy with life stuff, then got mentally blocked on wanting to get nightly builds working before doing any further bespoke work. but that's finally done!

Issue #666, opened 2021-12-05
Hang on Exit Mac X86


baconpaul:
OK so

Mac x86
Load Arturia Chorus
Push any sound into it and out of it so you can hear one chorused note
Quit

At this point bespoke hangs and it shows this in the log

```cpp
JUCE Assertion failure in juce_VST3PluginFormat.cpp:3432
JUCE Assertion failure in juce_VST3PluginFormat.cpp:3466
JUCE Assertion failure in juce_VST3PluginFormat.cpp:3432
JUCE Assertion failure in juce_VST3PluginFormat.cpp:3466
JUCE Assertion failure in juce_VST3PluginFormat.cpp:3432
JUCE Assertion failure in juce_VST3PluginFormat.cpp:3466
JUCE Assertion failure in juce_VST3PluginFormat.cpp:3432
JUCE Assertion failure in juce_VST3PluginFormat.cpp:3466
libc++abi: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
```

baconpaul:
Those asserts come from 'begin edit' / 'end edit' on the plugin

The leak detector also shows that we leak the instance of the VST3Plugin

PR #668, opened 2021-12-05
Cleanup the mPlugin on ::Exit

baconpaul:
The mPlugin is left around forever which means that the
plugin is not terminated on exit. With some plugins,
as outlined in #666


awwbees:
looks good!

PR #684, opened 2021-12-10
Added midi import and export to note canvas

gusmccallum:
Note canvas can now load and save .mid files with "load midi" and "save midi" buttons respectively.
Things to note:

- Any midi type 0 or 1 file can be loaded but only the first track will be loaded for multi track type 1 midi files, pending a UI solution to select an individual track to import    
- Transport tempo not adjusted by incoming midi file    
- Transport tempo not exported in outgoing midi files    
- Transport time signature not adjusted by incoming midi files    
- Transport time signature is exported in outgoing midi files

### DIFF
--- a/Source/NoteCanvas.cpp
+++ b/Source/NoteCanvas.cpp
@@ -36,6 +36,8 @@
 #include "CanvasTimeline.h"
 #include "CanvasScrollbar.h"
 
+#include "juce_gui_basics/juce_gui_basics.h"
+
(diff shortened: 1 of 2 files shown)


ScriptTiger:
I really love the premise of this feature, but I can't seem to load .mid files which were previously saved. The saved .mid files seem to load fine in other applications, like FFmpeg, Audacity, etc., but I just can't load them with the notecanvas.

PR #692, opened 2021-12-19
Support `hotloading` prefab title bar options

shiehn:
Issue: 
- currently the title bar options for prefabs are loaded at application load time.  If a prefab is removed from the file system and the Bespoke user selects the cached prefab the application will crash
- this enables the use case where a user adds prefabs to the file system during a Bespoke session and essentially "hot loads" them 

Solution:
- read the prefabs directory at runtime when the user selects the prefabs title bar drop down (DropdownClicked).  I believe this is a lightweight operation and improves both user experience and stability

### DIFF
--- a/Source/TitleBar.cpp
+++ b/Source/TitleBar.cpp
@@ -248,10 +248,7 @@ void SpawnListManager::SetModuleFactory(ModuleFactory* factory)
    mOtherModules.SetList(factory->GetSpawnableModules(kModuleType_Other), "");
    
    SetUpVstDropdown();
-   
-   std::vector<std::string> prefabs;
-   ModuleFactory::GetPrefabs(prefabs);
-   mPrefabs.SetList(prefabs, "prefab");
+   SetUpPrefabsDropdown();
    
    mDropdowns.push_back(&mInstrumentModules);
    mDropdowns.push_back(&mNoteModules);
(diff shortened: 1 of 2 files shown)


shiehn:
ps - manually tested on: Windows 11 & Ubuntu 20.04

awwbees:
looks good! thanks!

unfa:
Is it possible to hot load prefabs using a script? Say I want to load a different patch into a prefab that's my drumkit, but keep connections from approprietely named nodes (say "note_in", "control_in" and "audio_out").

I am worried my laptop won't be able to handle a massive Bespoke Synth project for a live set, so I am looking into ways to partition the set and have these parts loaded and discarded on demand.

Needless to say - it'd need to not interrupt the audio thread... Right now when I try loading a prefab during playback, this can stall the entire audio thread and cause a nasty stutter or "buzzzz" while a short audio buffer is repeated over and over.

NoxyNixie:
> Is it possible to hot load prefabs using a script? Say I want to load a different patch into a prefab that's my drumkit, but keep connections from approprietely named nodes (say "note_in", "control_in" and "audio_out").

This is not possible. You should make a new feature request (or bug report) for this since Awwbees rarely if ever looks at closed issues. Even if this was possible it will likely still interrupt the audio thread especially if you have `autosave` enabled.

PR #704, opened 2021-12-27
Add MPK mini 3 controller

mjy:
One thing I'm uncertain about is the use of `channel` related settings.  The pads are channel 10, everything else is 1.  Not sure if that needs to be defined in the file here.


awwbees:
thanks! I think I'll need to add functionality to midicontroller to handle these devices which send out on multiple channels like this. there isn't good support for it yet. I have a MPK midi 3, so I have a good way to test it.

one unrelated odd thing in here is that I see that `tuning-library` has some mysterious change. I assume that was an accidental inclusion in this PR?

PR #742, opened 2022-01-27
Make chorder emit notes from lowest to highest

sbungartz:
When using `chorder` with `strummer` or `arpeggiator` I noticed that while the notes within an octave are ordered from lowest to highest, the octaves themselves are ordered highest to lowest.
While this can obviously create unique effects of it's own, I would have expected it to emit notes in pitch order.

This PR changes the outer loop of iterating through the grid rows to go from bottom row to top row, so that all notes are ordered by pitch.

This can obviously quite drastically change the sound of existing patches, so this is a breaking change.
Another option would be to make the `chorder` more configurable regarding the output order.


awwbees:
makes sense to me, thanks for the fix!

PR #752, opened 2022-02-09
Add Python-Binding midicontroller.send_sysex

sbungartz:
Allows sending Sys-Ex messages to Midi-Devices to use Vendor-Specific features

### DIFF
--- a/Source/MidiController.cpp
+++ b/Source/MidiController.cpp
@@ -1526,6 +1526,14 @@ void MidiController::SendData(int page, unsigned char a, unsigned char b, unsign
    }
 }
 
+void MidiController::SendSysEx(int page, std::string data)
+{
+   if (page == mControllerPage)
+   {
+      mDevice.SendSysEx(data);
+   }
+}
+
 UIControlConnection* MidiController::GetConnectionForControl(MidiMessageType messageType, int control)
 {
    for (auto i=mConnections.begin(); i != mConnections.end(); ++i)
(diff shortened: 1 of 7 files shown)


unfa:
This would allow to for example use Launchpad Mini MK3's extra features like switching it into Programmer mode automatically, displaying scrolling text, changing LED brightness etc - this would be very useful in eliminating some manual steps required to get a live set going, and also would help in recovering form a general failure like an OS reboot etc.

I wonder if this could be worked around with Python somehow before this gets implemented?

sbungartz:
Hi @unfa, just to avoid any misunderstandings: This is implemented in the Nightly build.
I have only been using the nightly builds, because there is so much new stuff that has been added since the last release.

If you are using the nightly build, you can do `m.send_sysex(bytes([0, 32, 41, 2, 13, 0, 127]))` given a `midicontroller` in variable `m`. The bytes `240` and `247` are added to the beginning and end of the message by the `send_sysex` command, since they are what makes a MIDI message a sysex message.

If you do not want to use the nightly, I'm not sure how to do this otherwise. Might be possible if you can load a python midi library like [`mido`](https://mido.readthedocs.io/en/stable/) into the bespoke context, but I have not tried loading external libraries into bespoke before...

(Off topic: Thanks for introducing me to Bespoke in the first place over on YouTube 😊)

unfa:
Oh, that's great news! I'm going to upgrade then :D

PS: Glad I could help :)

PR #785, opened 2022-03-16
prototype - chunked and annotated output

shiehn:
**Status: WIP - not ready for merge**

**Use case:**
- user wants the bespoke output to be organized and annotated by x number of bars, key, tempo
- a common workflow for producers is based around pulling from loop collections.  In order for that to be effective the loops must be organized and annotated to match the target project

**Whats changed:**
- added an option to create an array of rolling buffer arrays to collect the output if chunks of specified bars

**Open Convo/Subject to change:**
- the output is currently dumping to `/BespokeSynth/recordings/chunked-output`
- the file format is currently [BPM]-[KEY]-[BarIndex]
- what should happen if a user writes to the output mid way through a chunk
- what should happen if a user writes to the output with multiple keys or tempo mid chunk


shiehn:
I'm closing this PR.  After chatting with some folks from the discord and testing this feature as a 'user' I now believe that a looper pedal style audio export is a better approach

PR #801, opened 2022-03-30
SynthGlobals: Switch from std::mt19937 to Xoshiro256**

modeco80:
this PR switches the global RNG to an implementation of Xoshiro256** implemented as a C++ random number engine.

Skipping past mt19937 being extremely easy to underseed (and the previous code doing so),
mt19937 also carries a huge amount of baggage state which makes generating random data take it's sweet time.

However, xoshiro256**:

 - has a fraction of the state size
 - cannot be underseeded as easily (if at all),
 - is simply faster for a tiny bit less entropy (well, considering underseeding, it might be *more* entropy.).

### DIFF
--- a/Source/SynthGlobals.cpp
+++ b/Source/SynthGlobals.cpp
@@ -79,7 +79,7 @@ bool gShowDevModules = false;
 float gCornerRoundness = 1;
 
 std::random_device gRandomDevice;
-std::mt19937 gRandom(gRandomDevice());
+bespoke::core::Xoshiro256ss gRandom(gRandomDevice());
 std::uniform_real_distribution<float> gRandom01(0.0f, 1.f);
 std::uniform_real_distribution<float> gRandomBipolarDist(-1.f, 1.f);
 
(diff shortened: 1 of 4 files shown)


modeco80:
alright, rebased and formatted. It should be ready now.

note: if this temporarily fails due to the force push being another push (therefore another workflow run?) it's my bad (i could have done the rebase + force push + commit in one push, but lazy)

awwbees:
is there any license information for xoshiro256**? I looked around but couldn't find anything.

modeco80:
The implementation I used is under public domain, and the original C version which *that* is based off of (linked in the orignal comments, which I kept in) has:

```cpp
/*  Written in 2018 by David Blackman and Sebastiano Vigna (vigna@acm.org)

To the extent possible under law, the author has dedicated all copyright
and related and neighboring rights to this software to the public domain
worldwide. This software is distributed without any warranty.

See <http://creativecommons.org/publicdomain/zero/1.0/>. */
``` 
which seems to mean it's also in the public domain.

PR #812, opened 2022-04-07
OSC fixes

NoxyNixie:
Fix that the OSC controller ignored any arguments beyond the first. This bug would make it impossible to use widgets like the `xy`, `range`, `rgb`, `multixy` and others.

Made it possible to receive note data through OSC using addresses that start with `/note`.  By default the notes will be seen as coming from channel 1 but if you use `preArgs` in Open Stage Control one can set the channel to something else by specifying an integer preArg indicating the channel like the following:
image
The above will send the notes to channel 4 which you then can filter with the midicontroller's channel filter (not that two osccontrollers modules work at the same time but I think that is a different issue).

### DIFF
--- a/Source/MidiController.cpp
+++ b/Source/MidiController.cpp
@@ -356,7 +356,7 @@ void MidiController::OnTransportAdvanced(float amount)
       {
          playTime = gTime + (note->mTimestampMs - firstNoteTimestampMs);
          if (playTime <= lastPlayTime)
-            playTime += .01f; //hack to handle note on/off in the same frame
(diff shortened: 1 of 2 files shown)


awwbees:
cool, I checked the multiparameter version and it looks good! I need to take a look at an open stage control tutorial to set up the note control to test it out (and take a look at the stuck notes issue you brought up on discord)

NoxyNixie:
Awesome, and sorry for forgetting to remove that one debug line ;)

PR #815, opened 2022-04-15
make the save button visible in the presets module

can-opener:
this just makes the save button visible and moves it to the right so it's not behind a slider. I tested it out with the presetsfile option in the triangle menu of the presets module and it seems to work.

### DIFF
--- a/Source/Presets.cpp
+++ b/Source/Presets.cpp
@@ -52,12 +52,12 @@ void Presets::CreateUIControls()
 {
    IDrawableModule::CreateUIControls();
    mGrid = new UIGrid("uigrid", 5, 38, 120, 50, 8, 3, this);
-   mSaveButton = new ClickButton(this, "save", 50, 3);
+   mSaveButton = new ClickButton(this, "save", 130, 3);
    mBlendTimeSlider = new FloatSlider(this, "blend ms", 5, 20, 70, 15, &mBlendTime, 0, 5000);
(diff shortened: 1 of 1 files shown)


NoxyNixie:
I'm not sure this functionality actually works as expected. The original disabling of the save button was ~14 months ago in 1f120f0bb80f5d8a171cfe9e14092d7a900c9060 and seems to indicate that it was done to avoid save corruption.

On top of that the actual functionality isn't that useful. Yes you can save the preset but it is only ever loaded when you load your savestate (at which point you could let the normal saving take care of this instead of the additional presetfile). So it is not an easy way to try out different presets.

Even if this was made into an easy way to try out presets there are quite a few issues to work out before that is even possible. First of all you'd need to account for reconnecting the preset module to something completely different or even loading a preset for a module that was saved for a completely different module. And then you'd also need to handle the individual connections the preset module can have.

Also what would happen if the presetsfile was removed or you shared your BSK to someone else who does not have your presetfiles (If my understanding of the code is correct it would crash (mostly because json access already happens outside of the large try/catch block))?

Based on what I've seen in the code and the commit history I suspect that the presetsfile feature is an abandoned incomplete feature that was disabled for good reasons.

If the presets module gets a hot swappable presetfile system I think it should be in the main module UI much like the `script` or `drumplayer` modules and it should probably use a folder structure that includes the module name that it is connected to so that one can't easily load presets from other modules. However this isn't fool proof either since one can easily preset an entire prefab module with its contents which in turn can be anything including presets themselfs.

So yeah, this presetsfile thing sounds cool but has _many_ issues that would need to be resolved.

awwbees:
yeah, the presets "save" functionality is old, half-baked, and was intentionally disabled

PR #818, opened 2022-04-18
Use vst's ID instead of file path.

KottV:
Hi,

this is a hack and proof of concept for using plugin's ids.
vstplugin titlebar dropdown is working
loading old presets is working
spawning vsts on-the-fly by typing is not working

I still  think there is a room for improvements, at now it uses uniqueid only (as a list array index), but preferable way would be use a whole `PluginDespcription`


.gitmodules
```diff
@@ -1,6 +1,3 @@
-[submodule "libs/JUCE"]
-	path = libs/JUCE
-	url = https://github.com/juce-framework/JUCE
```
NoxyNixie: I don't think this is intentional?

KottV: yep

resource/help_rus.txt
```diff
@@ -0,0 +1,62 @@
...
+        чтобы быть уверенным, что вы не упустите ни одного крутого момента.
+   * кнопка "save state": сохранить текущую компановку рабочего пространства в файл. комбинация клавиш: ctrl-s
+   * кнопка "load state": загрузить ранее сохраненные рабочие пространства. комбинация клавиш: ctrl-l
+
+горячие кнопки
+   * ctrl/command-'s': сохранить
+   * ctrl/command-'l': загрузить
+   * shift-'p': пауза обработки звука. это остановит весь аудиовыход. снова нажмите shift-'p', чтобы продолжить
+   * F2: переключить отображение всех элементов управления adsr в виде слайдеров 
+        (при перетаскавании патча от модуля модуляции в зону adsr переключается автоматически)
```
NoxyNixie, line 62: Shouldn't this be part of a different PR?

resource/tooltips_rus.txt
```diff
@@ -0,0 +1,1794 @@
+
```
NoxyNixie, line 1: Shouldn't this be part of a different PR?

resource/userdata_original/controllers/FCB1010_1.json
```diff
@@ -0,0 +1,25 @@
+{
```
NoxyNixie, line 1: Shouldn't this be part of a different PR?

resource/userdata_original/controllers/FCB1010_4.json
```diff
@@ -0,0 +1,58 @@
+{
```
NoxyNixie, line 1: Shouldn't this be part of a different PR?

resource/userdata_original/controllers/FCB1010_9.json
```diff
@@ -0,0 +1,124 @@
+{
```
NoxyNixie, line 1: Shouldn't this be part of a different PR?

resource/userdata_original/controllers/SL MkII MIDI 1.json
```diff
@@ -0,0 +1,109 @@
+{
```
NoxyNixie, line 1: Shouldn't this be part of a different PR?

resource/help_rus.txt
KottV: it should, perhaps I forget that I added it in my master branch

NoxyNixie (commented):
Not sure if this was an accidental thing or not but it seems you've added and removed more than your PR describes ;)

KottV:
I'll rework these changes in a separate PRs

PR #826, opened 2022-04-28
Add a setting that allows the disabling of the module hilights

NoxyNixie:
Someone wanted to disable these ;)


awwbees:
nice! I had this on a list somewhere but forgot about it

PR #827, opened 2022-04-28
Fix the next and previous sample buttons in the sampleplayer pads on windows. Fixes #791.

### LINKED ISSUE #791: New drumplayer buttons don't work correctly
In bespoke NIGHTLY 1.1.0 (Mar 19 2022 02:17:53) - Windows 10 - HEAD 4bc5002 the back and forward buttons seem to both load the first sample in its folder:

BespokeSynth_20220322_102448

This was added in the following update: 268d845240198523329b6777c601e3bcdadd7e3f: #787

NoxyNixie:
It seems the file iterations returns paths with backslashes (`\`) while most of bespoke only uses forward slashes (`/`). This change should fix that.


awwbees:
thanks!

PR #828, opened 2022-04-29
Fix LFO. Fixes #718.

### LINKED ISSUE #718: LFO with S&H waveform doesn't work correctly with intervals bigger than 1n and free rate
## Describe the bug

If I set an LFO with S&H waveform to intervals bigger than 1n, I get the noise waveform instead of random values in the set interval. If I set it to free rate, I get a straight line that doesn't change at all.

## Build info

bespoke 1.1.0 (Nov 16 2021 11:41:32) - Windows 10 - main bd41884b9

## Steps To Reproduce

- load an LFO
- set osc type to s&h
- set interval to 2 (or above) or to free

expected: straight line of random value that changes in the set interval
actual: interval > 1n = noise, interval:free = constant value

NoxyNixie:
- Add 32nt and 64n to the list of intervals.
- Hide the offset, shuffle and length sliders in drunk and S&H modes.
- Show the soften slider in S&H mode. Force length to 1 if in perlin mode so the visuals look correct.
- Fix S&H mode not working with all intervals.
- Fix and inverse the drunk modes distance so that faster transport makes bigger changes.

I still want to make the free running mode work correctly in S&H mode but don't know how to approach that at this time (still have a lot to learn about bespoke source).


NoxyNixie:
Alright I fixed free mode for both drunk and S&H now too. I think this is ready for review / merging!

awwbees:
I fixed up the issue from my above comment (so that dotted intervals aren't treated like triplets), the rest of this looks probably good but I'm gonna do some further playing around with it. the LFO sure does have a lot of different branching paths to test!

NoxyNixie:
hm, I was actually aware that dotted was calculated wrong and was intending to do that in a separate PR combined with other transport changes as those changes have far greater consequences.

I already have some code going but it is taking its time since I have to figure out what is the "right" way to do things. I'm not sure how I should share my changes as it currently is a mess and I was hoping to do that after the LFO was "fixed"ish.

awwbees:
thanks for all of this cleanup!! running the LFOs with dotted intervals is really fun and sounds great, I hadn't ever thought to add that.

PR #834, opened 2022-05-04
Preset cleanup.

NoxyNixie:
- Removed unused saving and loading of a json preset file.
- Resize mPresetCollection when the grid is resized if needed. 
- Made sure all variables are initialized and added some protection when loading values that could be saved uninitialized. 
- Stopped MSVC from grumbling over potential arithmetic overflows by doing explicit casting.

I'm almost positive the explicit casting is unnecessary since the values won't get high enough to overflow but I'd rather be save than sorry.

I don't actually know if the loading value range checks fix any loading issues currently around since I don't have a good way to see what is going on during load (yet) and I don't actually think I have a savestate that exposes this particular issue. What prompted me to touch this code however is that the assert (https://github.com/BespokeSynth/BespokeSynth/blob/main/Source/Presets.cpp#L588) during saving was triggered many times while testing things with a preset in debug mode.

### DIFF
--- a/Source/Presets.cpp
+++ b/Source/Presets.cpp
@@ -32,16 +32,6 @@
 std::vector<IUIControl*> Presets::sPresetHighlightControls;
 
 Presets::Presets()
-: mGrid(nullptr)
-, mDrawSetPresetsCountdown(0)
-, mBlending(false)
-, mBlendTime(0)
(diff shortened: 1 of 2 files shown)


awwbees:
just hit the bug that this change takes care of, and remembered that I needed to review this PR. thanks for the fix!

PR #838, opened 2022-05-10
Fix freeverb.

### LINKED ISSUE #836: Freeverb is mono
## Describe the bug

The Freeverb plug-in seems to be mono. Should be stereo.

## Build info

bespoke 1.1.0 (Nov 16 2021 11:41:32) - Windows 10 - main bd41884b9

NoxyNixie:
Change the range of the width slide to be 0-100 instead of 0-1.
Change the initial freeverb values to half dry/wet and half width (similar to how dampening is set to half).

This also fixes #836.


NoxyNixie:
No idea why the mac build was too slow (it timedout and not actually errored):
image

awwbees:
yeah, that failure was just the mac azure build doing its "sometimes I randomly fail" thing, re-running it fixed the issue

thanks for this range fix! I always wondered why freeverb sounded so flat and boring, this explains it

PR #842, opened 2022-05-18
Eliminating uninitialized variables

### LINKED ISSUE #173: Valuestreams buffer too short or lowest speed too slow?
image

Setting the speed to anything below `0.4` shows "the end of the buffer". Should the buffer be made longer? or the default minimum speed set to `0.4` (instead of `0.1`)?

NoxyNixie:
The main focus of this PR is:
- Attempt at eliminating any and all sources of potential and not so potential uninitialized variables and transition towards curly brace initialization.

With additional fixes because I encountered them:
- Fixed a few potential dereferencing null pointers.
- Removed some (but by far not all) unused variables and functions.
- Fixed minor bug in valuestream where the speed slider could be set smaller than the actual data stream (Fixes #173).
- Fixed an instance where a for loop was making unnecessary copies.
- Fixed an infinite loop in `ScriptModule::RunScript` if `lineStart` was anything other than `-1`.

Not all files have been transitioned to curly brace initialization yet since I am mainly focusing on the files that have reported uninitialized variables. I know probably 99% of them are initialized elsewhere but I also know that there are more than a few that cause bugs:

- https://discord.com/channels/846834873254805554/854199359913328671/974333292453195806
- https://discord.com/channels/846834873254805554/890239086466580571/903970417516089424

and probably many more that I don't know of (yet). I'd rather be safe than sorry so I opted to initialize as much as possible.

I've tested this patches quite a bit but I know I'm not perfect and fully expect mistakes or errors or wrong assumptions.

I really hope you like this one as it was quite the job ;)

(I do know some of the files I have edited are hidden or even disabled modules, but I'd rather do some extra work than have these potential errors pop up later.)

(I also know that not all files have been transitioned to the curly brace initialization yet but those were not popping up with warnings so I skipped them in this first pass.)


awwbees:
slowly working my way through this one, it's a doozy! note to self: I'm currently at `DrumPlayer.h`

Source/IDrawableModule.cpp
```diff
@@ -272,7 +272,7 @@ void IDrawableModule::DrawFrame(float w, float h, bool drawModule, float& titleB
    }
 
    if (PatchCable::sActivePatchCable &&
-       (PatchCable::sActivePatchCable->GetConnectionType() != kConnectionType_Modulator || PatchCable::sActivePatchCable->GetConnectionType() != kConnectionType_UIControl) &&
+       (PatchCable::sActivePatchCable->GetConnectionType() != kConnectionType_Modulator || PatchCable::sActivePatchCable->GetConnectionType() != kConnectionType_UIControl) && //@TODO(Noxy): This is always true, what was the desired effect? (Also C6289 warning)
```
awwbees: oops! yep, that || should be a &&, fixed

awwbees:
note to self: now at `Polyrhythms.cpp`

Source/SamplerGrid.h
```diff
@@ -40,7 +40,7 @@
 
 class ofxJSONElement;
 
-#define MAX_SAMPLER_GRID_LENGTH 5 * 44100
+#define MAX_SAMPLER_GRID_LENGTH 5 * 44100 //@TODO(Noxy): Shouldn't this 44100 be gSampleRate instead?
```
awwbees: nah, it's just intended to be a constant-sized decent chunk of samples, 5 seconds-ish

Source/ModularSynth.cpp
```diff
@@ -2068,14 +2068,14 @@ void ModularSynth::ResetLayout()
...
 
-   TitleBar* titleBar = new TitleBar();
-   titleBar->SetPosition(0, 0);
-   titleBar->SetName("titlebar");
-   titleBar->SetTypeName("titlebar");
-   titleBar->CreateUIControls();
-   titleBar->SetModuleFactory(&mModuleFactory);
-   titleBar->Init();
-   mUILayerModuleContainer.AddModule(titleBar);
+   TheTitleBar = new TitleBar();
```
awwbees: why did you change how TheTitleBar is assigned? did it fix a warning?

NoxyNixie: The local variable is completely pointless there as far as I can see and the change actually does not change the program flow at all.

awwbees: gotcha. it makes more sense to me for singletons to set the global themselves (`TheTitleBar = this;`), so I'll switch it back

awwbees:
alright, let's get this in before any future merge conflicts make it too painful!!

thank you for all of this work! I looked over it all and everything seemed to line up. but obviously there's a such a large volume that something could have slipped through. fingers crossed!

PR #847, opened 2022-05-28
use gestures to detect when VST controls have been touched

### LINKED ISSUE #682: vst-plugin: `show parameter` doesn't respond to `parameter change request`
## Describe the bug

It is described in the mouse-over help that the `show parameter` option for the vst-plugin host should respond to parameter changes, however when a plugin sends a `parameter change request` it still doesn't show up.

Tested with: latest Cardinal build -> https://github.com/DISTRHO/Cardinal/commit/28d3ca2d689804b6acd7a07667a8ef6174e18407

Here it is simply sending this call, but with the current value. Perhaps BespokeSynth is actually expecting a _change_ in value? This shouldn't be needed though.

In this case the parameter is a pure input and doesn't have any controls so any value change would then have to be simulated.

## Build info

bespoke 1.1.0 (Dec  9 2021 21:41:07) - Linux - main 8c3f5c2

## Steps To Reproduce

Add plugin. Have plugin send `parameter change request` (connect one of the jacks inside). Do not see parameter appear in the list.

See the parameter appear in the list.

awwbees:
rather than value changes, which was my old naive way. fixes #682

### DIFF
--- a/Source/VSTPlugin.cpp
+++ b/Source/VSTPlugin.cpp
@@ -326,6 +326,7 @@ void VSTPlugin::LoadVST(juce::PluginDescription desc)
    {
       mPlugin->enableAllBuses();
       mPlugin->disableNonMainBuses();
+      mPlugin->addListener(this);
 
       /*
        * For now, since Bespoke is at best stereo in stereo out,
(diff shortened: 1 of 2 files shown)


baconpaul:
Cool!

PR #880, opened 2022-07-18
Add <memory> include

gnapiorkowski:
This missing header caused me an error when building on Linux. Adding it solved the problem.


awwbees:
thanks for the fix! #876 also includes this fix, which I'll be merging soon, so I'll hold off on this one to avoid the merge conflict

PR #883, opened 2022-07-21
Add TAB key to launch quick spawn

baconpaul:
Seriously, go into the canvas, do Tab/Surge/Enter. It's fun!


baconpaul:
So tab is cool but is used for other things but F2 seems safe!

PR #884, opened 2022-07-22
Create plugin parameter sliders on demand

KottV:
Previously sliders were created for all plugin's parameters, it took a long for plugins with enormous amount of them. We just didn't display these sliders.
This PR makes adding parameter (if them over 30) only on gesture then show it's slider.


awwbees:
VSTs with more than 31 parameters don't load savestates properly for me, so this needs some additional work before merging (but, it might not be very straightforward work)

KottV:
> VSTs with more than 31 parameters don't load savestates properly for me, so this needs some additional work before merging (but, it might not be very straightforward work)

it has to be deal with the savestate revisions afaik

awwbees:
I believe #896 fixed the root symptom here

PR #893, opened 2022-08-04
Added an option to have velocitytocv output 0 when a note is released.

flarn2006:
I was surprised this feature didn't already exist, and it seemed simple enough, so I added it. Please let me know if any further changes are needed.


awwbees:
sweet, thanks for the feature!

PR #894, opened 2022-08-07
Add boundstopulse module

BlockBuilder57:
boundstopulse creates a pulse when its input slider hits the min or max value.

I personally used this with a notetoggle input to make two separate pulses with the on and off note signal for a [walkie-talkie beep sound](https://user-images.githubusercontent.com/15617441/183276492-a3236e55-2df9-4324-a0ea-964e6568b52c.mp4), but this may also be able to find uses with LFOs and noise gates to toggle various things when those are hit/released.


awwbees:
sweet, neat module!

PR #895, opened 2022-08-07
Add Nektar Panorama P4 controller maps

BlockBuilder57:
The same layout should work for the P6, the only difference is key count.


awwbees:
thank you!

PR #909, opened 2022-08-29
Added an option to disable the 30-parameter limit for plugins

flarn2006:
Requiring the user to adjust a parameter to have it appear in the list is a good way to keep it clean, but what about plugins with parameters that aren't directly adjustable in the UI? I've added a setting (`vst_show_all_params`) which disables the 30-parameter limit, so those parameters are usable in Bespoke.

### DIFF
--- a/Source/BoundsToPulse.cpp
+++ b/Source/BoundsToPulse.cpp
@@ -1,0 +1,80 @@
+/**
+    bespoke synth, a software modular synthesizer
+    Copyright (C) 2022 Ryan Challinor (contact: awwbees@gmail.com)
+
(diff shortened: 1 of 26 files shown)


flarn2006:
For some reason the velocitytocv commit from #893 (which was already merged) is also listed under this pull request. I'm not sure why that is, but it doesn't appear to make a difference.

flarn2006:
I'm closing this because 1) recent changes have solved this issue in a better way, and 2) I forgot to create a separate branch for this and would like to be able to make unrelated commits to the branch. If the devs still feel like this would be a good addition, commit bb6194e is the one you want.

PR #914, opened 2022-09-09
Transition curly brace

NoxyNixie:
This is the long awaited curly brace initialization transition ;)

In theory this should not change much if anything at all.

I've tried to squash this stuff but it seems whatever I did merging in the new stuff made squashing cause conflicts. I have no idea how to resolve this.

Another issue is that I had some issues with the JUCE submodule which likely isn't the right revision and once again I have no idea how to resolve this correctly.

I hope this isn't too messy and if you want me to update this after other branches have been merged (there are likely conflicts) I'll happily fix whatever is needed ;)


awwbees:
sweet, thank you! I will hold off on merging this until after #904, since they both touch a huge number of files and there will likely be some conflicts

Source/ModuleContainer.cpp
```diff
@@ -125,13 +125,10 @@ void ModuleContainer::Clear()
    std::vector<IDrawableModule*> modulesToDelete = mModules;
    for (auto* module : modulesToDelete)
    {
-      if (module)
```
awwbees: why was this null check removed?

NoxyNixie: That is a mistake, good catch.

I must have had debug code in there at one point and probably removed a little too much when cleaning up.

NoxyNixie:
> I will hold off on merging this until after https://github.com/BespokeSynth/BespokeSynth/pull/904, since they both touch a huge number of files and there will likely be some conflicts

Understood and actually expected ;)

Source/Canvas.h
```diff
@@ -153,38 +153,38 @@ class Canvas : public IUIControl
...
+   bool mDragSelecting{ false };
    ofRectangle mDragSelectRect;
-   bool mDragCanvasMoving;
-   bool mDragCanvasZooming;
+   bool mDragCanvasMoving{ false };
+   bool mDragCanvasZooming{ false };
    ofVec2f mDragCanvasStartMousePos;
    ofVec2f mDragCanvasStartCanvasPos;
    ofVec2f mDragZoomStartDimensions;
    HighlightEnd mHighlightEnd;
```
awwbees: I was scanning through this PR and see that mHighlightEnd is now no longer initialized, but it was before this change. this one instance is easy enough to fix, but hmm, it makes me wonder if there's a foolproof way to scan over this and be sure that this large change won't introduce any other errors. I should probably write a python script to parse it all and identify any inconsistencies like this

NoxyNixie: Aight I see what I missed.

I think I can probably see those pop up if I do a code analyze run in MSVC.

NoxyNixie: Just did an analyze and it is not caught by MSVC which leads me to believe that these values are default initialized because they are an enum.

Since both were set to `kHighlightEnd_None` (in the original constructor) which is the integral value `0`, the code should be completely unchanged.

However I do see the value of explicitly initializing them for readability and potential other behavior with different compilers.

Up to you I guess if you want these explicitly initialized?

awwbees: yeah, they should probably be initialized. but my latter point is that I should put together some way of validating via a parser that this PR doesn't actually change any behavior, since I don't trust my eyes to not miss anything.

awwbees: ok! I wrote that parser, found a few minor discrepancies, and fixed them up. should be good to go now, with much greater confidence that this PR will have no side effects!

PR #916, opened 2022-09-11
Save / Load fixes.

NoxyNixie:
- Fixed a saving bug were the name was used instead of the type for child modules.
- Added a hacky workaround to fix the loading of savestates were child modules were saved with their name instead of type.
- Fixed crashes where the loading of a corrupted save would attempt to load invalid control or module names.
- Fixed loading crash where the end of the file is reached unexpectedly.
- Don't crash out when child count is incorrect in an attempt to load more.


Source/IDrawableModule.cpp
```diff
@@ -1133,7 +1140,7 @@ void IDrawableModule::SaveState(FileStreamOut& out)
 
    for (auto* child : mChildren)
    {
-      out << std::string(child->Name());
+      out << std::string(child->GetTypeName());
```
awwbees: hmm? I'm confused. I do in fact want to be saving the child's name, not the type name, so that LoadState() can find the child via its unique name. can you describe the issue that this is intended to fix? (or, is there an existing issue number?)

NoxyNixie: Honestly I don't remember (it's been over 3 months since I last looked at this).

Since this branch does not have the logging and it has been so long I don't remember why I changed this. I think it was because somewhere the type and name were mixed up but if I look at the code now I think you are right that this should be the name.

NoxyNixie:
I'm going to close that as there are conflicts and this probably needs a complete redo.

PR #917, opened 2022-09-12
security fix: allow users to review python scripts before running them

### LINKED ISSUE #911: .bsk files can run unsandboxed Python code without prompting the user (security risk)
## Describe the bug

If you have a script node in a .bsk file set to execute on init, the script will run automatically without prompting the user. However, the scripts don't appear to be sandboxed in any way; doing things like writing to the file system or executing shell commands is trivial. This makes .bsk files roughly equivalent to .exe files from a security standpoint, which is dangerous because even security-conscious users won't necessarily think of DAW files as something that can execute malicious code.

## Build info

bespoke 1.1.0 (Dec 20 2021 22:35:07) - Linux - makepkg bd41884

## Steps To Reproduce

1. Download the attached .bsk file and open it. [calculator.zip](https://github.com/BespokeSynth/BespokeSynth/files/9480411/calculator.zip)

*or*

1. Add a script module to a new patch.
2. Paste the following code into the module:
```python
import platform
from os import system

ostype = platform.system()
if ostype == "Windows":
   system("calc")
elif ostype == "Darwin":
   system("open /System/Applications/Calculator.app")
else:
   for cmd in ("xcalc", "gcalc", "kcalc"):
      if system(cmd) == 0:
         break
```
3. Save the patch, then close Bespoke, and reopen the file.

In either case, your system's calculator app will open. *(Note: the .bsk file won't work on a Mac since I forgot part of the path)*

My expectation is that either *a)* Bespoke will ask me before running the script, or *b)* the script will be prevented from accessing anything outside of Bespoke. However, neither of these were the case; the script immediately ran with full privileges.

One potential solution could be to sandbox scripts set to "execute_on_init", but not other scripts. This way, code that interacts with the system can still be used, but it won't be run unless the user specifically clicks the button to run it.

awwbees:
to protect users who download and open BSKs and prefab files from theoretical malicious users. until all scripts in a file are marked as trusted, no python scripts execute. trusted scripts are stored, so users only need to approve them once.

fixes #911

### DIFF
--- a/Source/ScriptModule.cpp
+++ b/Source/ScriptModule.cpp
@@ -48,6 +48,7 @@
 #include "pybind11/embed.h"
 #include "pybind11/stl.h"
 #include "leathers/pop"
+#include "juce_cryptography/juce_cryptography.h"
 
 namespace py = pybind11;
 using namespace juce;
@@ -79,6 +80,8 @@ true;
 #else
 false;
 #endif
+//static
+bool ScriptModule::sHasLoadedUntrustedScript = false;
 
 //static
 ofxJSONElement ScriptModule::sStyleJSON;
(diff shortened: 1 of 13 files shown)


Source/ScriptModule.cpp
```diff
@@ -1511,6 +1578,40 @@ void ScriptModule::LoadState(FileStreamIn& in, int rev)
...
+
+   if (!isScriptTrusted)
+      sHasLoadedUntrustedScript = true;
+   mIsScriptUntrusted = !isScriptTrusted;
+}
+
+juce::String ScriptModule::GetScriptChecksum() const
+{
+   juce::String code = mCodeEntry->GetText(false);
+   juce::String checksum = juce::SHA256(code.toUTF8()).toHexString();
```
NoxyNixie, line 1604: Good you are not using crc32 or md5!

awwbees: I used MD5 at first, then did some googling and switched to SHA256

PR #929, opened 2022-09-30
Fix some thread deadlocks with VST window creation

baconpaul:
VCV Rack Pro would deadlock on creation of the UI window in bespoke since 3e1d28441. There are two thigns to fix this

1. Upgrade JUCE 7.0.1 to 7.0.2 and
2. Make sure to create the UI on the correct thread and not interleaved with a modal polling event by having the window creation done in a juce::Timer::callback scheduled onto ui thread.

### DIFF
--- a/Source/VSTPlugin.cpp
+++ b/Source/VSTPlugin.cpp
@@ -597,9 +597,12 @@ void VSTPlugin::Poll()
    {
       if (mPlugin != nullptr)
       {
-         if (mWindow == nullptr)
-            mWindow = std::unique_ptr<VSTWindow>(VSTWindow::CreateVSTWindow(this, VSTWindow::Normal));
-         mWindow->ShowWindow();
+         juce::Timer::callAfterDelay(0, [this]() {
+                                        if (mWindow == nullptr)
+                                           mWindow = std::unique_ptr<VSTWindow>(VSTWindow::CreateVSTWindow(this, VSTWindow::Normal));
+
+                                        mWindow->ShowWindow();
+                                     });
 
          //if (mWindow->GetNSViewComponent())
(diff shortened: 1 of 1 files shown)


baconpaul:
So this doesn't *quite* fix Rack. It fixes it about 4 times in 5 for me (whereas without it it reliably deadlocks every single time).

If you get it deadlocked what you will find is that theres a juce timer which calls to rack idle which calls to glfwJoystickPresent which calls to a CFRunLoop which never gets serviced because we are already in the UI queue run loop.

Screen Shot 2022-09-30 at 10 47 09 AM

I wonder if there is some terrible juce open gl initialization we have to do that we are missing somehow? That's the next thing that comes to mind.

But clearly the combo of the commit plus move to juce 7 has lead to a creation threading bug which blows up rack pro in bespoke but not in the audio plugin host?

Ugh.

awwbees:
in your opinion, does this change still seem like a good thing to merge, even though it doesn't fully fix this specific VCV issue? it does seem like this change is the correct thing to do generally.

baconpaul:
Well 702 for sure 

the try later is because I originally thought the poll may not be on the main thread where uis like to be created but now I think it just masked the problem. This would guarantee main thread creation if that poll ever changes though so “can’t hurt doesn’t fix” is indeed my conclusion 

whatever combo of OpenGL juce7 and timing is making us blow up …. I almost just wanted to do a glfw joystick init in bespoke main and pretend the problem doesn’t exist but we don’t actually bind glfw :)

baconpaul:
This was correctly fixed by 5a3eb8e2e9c71e2d2e26c67be77ab678bcdc8219

we probably want a 702 upgrade separately but this can close.

PR #931, opened 2022-10-03
Fix a Linux clang build error

ghost:
Building with the clang compiler on Linux errors here,
with `sqrt()` returning a double.
```
error : non-constant-expression cannot be narrowed from type 'double' to 'float' in initializer list
```

### DIFF
--- a/Source/BiquadFilter.h
+++ b/Source/BiquadFilter.h
@@ -68,7 +68,7 @@ public:
    void Filter(float* buffer, int bufferSize);
 
    float mF{ 4000 };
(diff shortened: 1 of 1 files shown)


awwbees:
thanks! irritating that my CI did not catch this

haenkel:
ci couldn't have caught this, it uses gcc for linux which doesn't complain about this one 
and apple-clang and linux-clang are also different enough it seems. 
You could consider using clang for the incoming debug build and gcc for release (or the other way round)
if you want to be safe(r), 
but then again it's probably not super often that things error on one and don't even warn on the other.

PR #936, opened 2022-10-06
Handle param rename events from plugins

baconpaul:
Parameter renames create a changed event, so use that event to make the subsequent poll rename the associated sliders. This makes VCVRack Pro basically usable for automation but I bet it also fixes the problems with F-Em and others.

One change: Since the rack parameter names are quite long (things like "Attack (Bog Audio ADSR module)") I changed the param name query from 32 to 64 characters in the newly introduced get-name function.

### DIFF
--- a/Source/Slider.cpp
+++ b/Source/Slider.cpp
@@ -197,7 +197,7 @@ void FloatSlider::Render()
    else
    {
       if (mShowName)
-         display = std::string(Name());
+         display = std::string(mHasOverrideDisplayName ? mOverrideDisplayName : Name());
       if (display.length() > 0) //only show a colon if there's a label
          display += ":";
       if (mFloatEntry)
--- a/Source/Slider.h
+++ b/Source/Slider.h
@@ -92,6 +92,11 @@ public:
    float GetModulationRangeMin() const override { return mMin; }
    float GetModulationRangeMax() const override { return mMax; }
    void OnTransportAdvanced(float amount) override;
+   void SetOverrideDisplayName(std::string name)
+   {
+      mHasOverrideDisplayName = true;
+      mOverrideDisplayName = name;
+   }
 
    void Init() override;
 
@@ -185,6 +190,8 @@ private:
    int mLastComputeSamplesIn{ 0 };
    double* mLastComputeCacheTime;
    float* mLastComputeCacheValue;
+   bool mHasOverrideDisplayName{ false };
+   std::string mOverrideDisplayName{ "" };
 
    float mLastDisplayedValue{ std::numeric_limits<float>::max() };
 
--- a/Source/VSTPlugin.cpp
+++ b/Source/VSTPlugin.cpp
@@ -509,6 +509,28 @@ bool VSTPlugin::ParameterNameExists(std::string name, int checkUntilIndex) const
    return false;
 }
 
+const std::string VSTPlugin::getUniquifiedParameterName(int i, const juce::Array<juce::AudioProcessorParameter*>& parameters)
+{
+   if (i < 0 || i >= parameters.size())
+   {
+      return std::string("ERROR PARAM ") + std::to_string(i);
+   }
+   auto originalParamName = parameters[i]->getName(64).toStdString();
+   auto name = originalParamName;
+   try
+   {
+      int append = i;
+      while (ParameterNameExists(name, i) || FindUIControl(name.c_str()))
+      {
+         ++append;
+         name = originalParamName + std::to_string(append);
+      }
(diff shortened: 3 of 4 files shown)


awwbees:
there's something in here to still be sorted through: it doesn't seem to handle savestates correctly yet. it looks like when `setCurrentProgramStateInformation()` is called, `audioProcessorChanged()` doesn't get called until after an async update. but, `VSTPlugin::LoadState()` calls `setCurrentProgramStateInformation()` and then expects to immediately be able to reference the parameters by name to load them. so, in VCV rack for example, it's expecting to reference something like "Attack (VCV ADSR)" at load, but the parameter is still called "Unassigned #2" at this time as far as bespoke knows.

haven't thought too hard yet, but three potential solutions I see:
1. find a way to get the VST parameter names immediately after calling `setCurrentProgramStateInformation()`
2. make the sliders keep the original names the VST reports under the hood for the purposes of bespoke addressing them, and just update the displayed names that the user sees. this will surely create a few holes here and there, but is probably acceptable
3. massively refactor how bespoke loads state and/or addresses UI controls (doubtful that I'm gonna do this one)

baconpaul:
> there's something in here to still be sorted through: it doesn't seem to handle savestates correctly yet. it looks like when `setCurrentProgramStateInformation()` is called, `audioProcessorChanged()` doesn't get called until after an async update. but, `VSTPlugin::LoadState()` calls `setCurrentProgramStateInformation()` and then expects to immediately be able to reference the parameters by name to load them. so, in VCV rack for example, it's expecting to reference something like "Attack (VCV ADSR)" at load, but the parameter is still called "Unassigned #2" at this time as far as bespoke knows.
> 
> haven't thought too hard yet, but three potential solutions I see:
> 
> 1. find a way to get the VST parameter names immediately after calling `setCurrentProgramStateInformation()`
> 2. make the sliders keep the original names the VST reports under the hood for the purposes of bespoke addressing them, and just update the displayed names that the user sees. this will surely create a few holes here and there, but is probably acceptable
> 3. massively refactor how bespoke loads state and/or addresses UI controls (doubtful that I'm gonna do this one)

ugh right

so the problem here is: we should never ever index or persist vst parameter names. For VST2 use the parameter index and for VST3 AU and CLAP use the parameter ID. 

But it is probably a bit late for that.... 

the problem is an upgrade of rack could change "Untitled (3)" to "To Be Set (3)" and every bespoke session would break. Or, we rename surge parameters all the time. etc...

tricky.

awwbees:
ok yeah, so it sounds like the solution for naming sliders is a variant on #2, but instead of using the first reported name, we instead use the parameter index (for vst2) or parameter ID (for other plugins). and then the sliders just draw with a special display name override, rather than showing their actual under-the-hood name.

awwbees:
ok, I just pushed the solution I proposed above. let me know if it makes sense on your end, from the plugin-writer perspective. the sliders are now named "paramid_[parameter id]" (or "param_[index]" for VST2), and they get a fake display name to make them readable to users.

the downside is that in the few places that slider names are exposed to users (`midicontroller`'s mapping, or addressing sliders in `script`), users now have to use the generic "paramid_[parameter id]" names to address them. but, the upside is that their save files will continue to work if the author of the plugins they're using decides to rename something. so that seems like a fair trade.

there's also a bit of extra garbage in here to retain the old way of naming sliders, so old save files continue to work.

KottV:
F-em has 47137 parameters. I had to adjust MIN of `numParameters` to 48000 to make it work. Otherwise seems fine :)

awwbees:
@KottV oh! you just made me realize that as a side effect of this change, we can now remove that `numParameters` cap entirely. cool!

awwbees:
I'm going for it! lemme know if you notice any gaps in this approach @baconpaul

baconpaul:
Oh awesome! Lemme pull it down and take a look in surge and rack :)

baconpaul:
Looks great in a 10 second test!

At some point in the future, maybe we want to use the plugin parameter -> string method in the display once you get a control but that's separate. 

Thanks for fixing and merging :)

awwbees:
are you referring to `getCurrentValueAsText()` and `getAllValueStrings()`? yeah, I first noticed those while making this change! I didn't realize this type of information was exposed to the host. that would be a great feature to add to bespoke's plugin hosting.

baconpaul:
yes exactly but that's a separate issue.
but its how, say, surge shows filter frequency in bitwig and stuff in param controls. the api is all there we just need to plumb it to the slider somehow.

Issue #937, opened 2022-10-09
macroslider does not work with inputs larger the output's inferred max


JustAPerson:
## Describe the bug

Here's a video demonstrating where adding a macroslider breaks a working modulator setup

https://user-images.githubusercontent.com/789308/194771655-7c6d3e75-c1c0-4b9e-8d81-f6e653f3b19a.mp4

In the beginning, there is an expression module whose output peaks at 8192. This is able to drive the `freq` param of a signalgenerator even though that has a default max of 4khz. Once I add a macroslider on the output of the expression module, things break and the signalgenerator is no longer being modulated.

## Build info

`bespoke 1.1.0 (Oct  8 2022 14:30:30) - Mac OSX 12.3.1 - main 616ad67c`

## Steps To Reproduce

- Create an `lfo`, `expression`, and `signalgenerator` modules
- Set the lfo to sweep from 0 to 1 and feed it into the input of the expression module
- Set the expression module to evaluate `2^(7+6*x)` and feed the input into the `freq` param of the signalgenerator
- Shift-left click the output of the expression module to add a macroslider
- Observe that the signalgenerator is no longer being modulated

## Expected vs. actual behavior

Honestly I just wish shift clicking a modulator did the exact same as every other type of node. I want to literally duplicate the output of the modulator. I do not want the extra functionality of the `macroslider` whatsoever.

awwbees:
I've got a PR at https://github.com/BespokeSynth/BespokeSynth/pull/944 which allows you to duplicate the output of a modulator as desired. I need to test it more before merging it, but if you're able to build code, you can help me test it by grabbing that branch

KottV:
@awwbees @JustAPerson Can we closed it as fixed with https://github.com/BespokeSynth/BespokeSynth/pull/944 ?

NoxyNixie:
I tested the originally presented issue with the latest Nightly build and can confirm that this has been resolved:

![BespokeSynth_20221029_131636](https://user-images.githubusercontent.com/211381/198828502-30018f03-a989-4c43-9174-5d4071510f2c.gif)

PR #944, opened 2022-10-21
allow modulators to directly connect to multiple UI controls, no macroslider necessary

awwbees:


awwbees:
resolves #937

NoxyNixie:
Tried and looks good so far!

Both the direct connected and through `macroslider` behavior seems to be much better and self explanatory.

image

The only thing I've noticed so far is that the direct connection on an `ADSR` visually behaves differently in comparison to a connection through the `macroslider` and leaves the cable in a location that is not directly visually linked to that `ADSR`.

Below a small savestate that demonstrates this oddness but also tests audibly if there is a difference between the two ways of modulation and so far my ears can't really tell when the `lfo` switches from using the two or just the one `fmsynth` (I'd say this is a good thing ;) ).

Here is the savestate in case you want to try too:
[2022-10-21_16-50.zip](https://github.com/BespokeSynth/BespokeSynth/files/9840209/2022-10-21_16-50.zip)

awwbees:
ah yeah, that is an exposure of a slight oddness that probably needs to settle out someday. there are secretly two types of modulation connections: "true" modulators, and value setters. "true" modulators lock out the targeted slider and completely control it. value setters just have values "pushed" from the modulator, but the slider can otherwise be controlled.

generally, value setters are for instantaneous things (such as the `controlsequencer`), and true modulators are for continuous things (such as the `lfo`). you can see a subtle visual difference: I made value setter cables slightly greyer than true modulator cables. maybe some day I'll clean it up and find a way to collapse these different types together, but for now, they're subtly different.

one specific thing here is that it does seem like `curvelooper` should probably be a true modulator instead of just a value setter, I'll take a look at that

PR #980, opened 2022-12-07
Fill up the dropdown so that all preset values will be selected when modulating the preset selector.

NoxyNixie:


awwbees:
the original implementation here was intentional, not an oversight. I don't want to default to making the dropdown enormous to accommodate the edge case of a user who wants 400 different presets

PR #1053, opened 2023-01-12
Allow the OSC control messages to work in percentage mode.

NoxyNixie:


awwbees:
sorry that I haven't merged this one yet... I'm a bit hung up on two things in the first commit:
- the normalized value seems better suited as being in the 0-1 range instead of a 0-100 percentage
- adding a "1" to the end seems like a confusing interface, and possibly more difficult to set up in some OSC interface software (which might prefer to just send messages in the format "<address> <value>"). maybe instead it should be a separate address, like /bespoke/control_normalized or bespoke/normalized (although I'm not super happy with either of those options, so probably a better name, which I haven't thought of yet)

and in the second commit, what is the purpose of making /note and /bespoke/note both valid addresses? I don't quite follow the feature being added here.

NoxyNixie:
> * the normalized value seems better suited as being in the 0-1 range instead of a 0-100 percentage

I can do between 0 and 1. I was initially thinking it is a percentage and humans tend to express those as 0 to 100. But honestly I'm indifferent about which one is used.

> * adding a "1" to the end seems like a confusing interface, and possibly more difficult to set up in some OSC interface software (which might prefer to just send messages in the format " "). maybe instead it should be a separate address, like /bespoke/control_normalized or bespoke/normalized (although I'm not super happy with either of those options, so probably a better name, which I haven't thought of yet)

What about `/bespoke/clamped` or `/bespoke/clamped_control`?

> and in the second commit, what is the purpose of making /note and /bespoke/note both valid addresses? I don't quite follow the feature being added here.

It is to make it possible to have Bespoke talk to Bespoke over OSC using notes. The `oscoutput` module sends notes out at address `/bespoke/note` instead of just `/note`. So I just added `/bespoke/note` as addition. Though one could argue that only having `/bespoke/note` be valid is better but that would break mappings of people who already use `/note` (though this hasn't been in the wild that long so could probably be done with minimal backlash).

NoxyNixie:
> * the normalized value seems better suited as being in the 0-1 range instead of a 0-100 percentage

Owh just thought of this. I also did it to support sending `int`'s as percentage. Granted you'd still be limited in resolution if you use ints but at least you won't be limited to just 0% and 100% and nothing in between.

NoxyNixie:
Okay, I did some work but I'm not sure this is all 100% since I'm sleep deprived.

But I think it all works now (at least in my testing everything seems to be good).

Commit message:
> Added character escaping to the address portion of OSC messages (and appropriate decoding).
> Changed the percentage mode to use a separate address instead of an additional argument.
> Resolved issues with controlling drop down lists.

PR #1079, opened 2023-01-23
Fix #1076 (Bug - can't pickup cables that are inside prefabs).

### LINKED ISSUE #1076: Bug - can't pickup cables that are inside prefabs
(Re-opened due to reasons mentioned in [this comment](https://github.com/BespokeSynth/BespokeSynth/issues/1076#issuecomment-1611054718).)

## Describe the bug

In certain cases, it is impossible to grab a cable.

## Build info

commit b0e179b5a6d5c6433965d92a943bb1019187dc15

## Steps To Reproduce

1. Make this effect chain

2. Try to grab any of the cables by either end (it should be impossible)

However, you should be able to grab the audio cable if you grab it by the small bit of the blue circle that is inside the prefab.

Cables should be able to be grabbed by either end, regardless of whether they're inside a prefab or not. Currently, cable ends cannot be grabbed while inside prefabs, which makes it impossible to remove cables from a `presets` module if it is inside a prefab. Additionally, cable beginnings cannot be grabbed if they stick out of a prefab (as shown above).

NoxyNixie:
I do have to admit that this is a workaround rather than a proper fix. Not sure how to tackle this properly though.

### DIFF
--- a/Source/Prefab.cpp
+++ b/Source/Prefab.cpp
@@ -65,13 +65,18 @@ std::string Prefab::GetTitleLabel() const
    return "prefab";
 }
 
+namespace
+{
+   const int kPatchCableSourceClickRadius = 7; // Not sure how to reference the actual value in PatchCableSource.cpp here.
+}
+
 void Prefab::Poll()
 {
    float xMin, yMin;
    GetPosition(xMin, yMin);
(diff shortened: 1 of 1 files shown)


awwbees:
resolved by #1148

PR #1085, opened 2023-01-24
Allow multiple instances of sampleplayers to download different youtube videos at the same time.

NoxyNixie:
I know this isn't entirely fool proof since in theory people can still have the same named sampleplayer (i.e. in a prefab).

Maybe I should use the full module path instead?


NoxyNixie:
hmmm what if I just use the address of the sampleplayer instance ...

NoxyNixie:
That last commit no longer relies on the sampleplayer name but instead the sampleplayer instance's memory address which should be perfectly fine for a temporary download of audio.

awwbees:
seems like a fine solution, although it exposes an existing issue: I never delete the downloaded youtube.wav, which before wasn't great and meant you'd always have youtube.wav sticking around but at least you'd only have one, but now they'd accumulate. I need to add some code to clean up the files after the youtube download is complete.

PR #1089, opened 2023-01-24
Notegate fix. Resolves: #163.

### LINKED ISSUE #163: notegate can fail to turn off notes
Here's a notegate that lets through C notes with velocity 127:

gatefail1

Here's the same gate turned off:

gatefail2

The script node shows me that the moment I turned the gate off, it sent a C note with velocity 0, then no more notes. The oscillator is off now.

This situation looks almost identical:

gatefail3

Again, I shut the gate and the script node shows a note-off event and no further activity. However, the cables to the script node and oscillator now show steady activity, and the oscillator keeps playing the last note instead of going silent.

I haven't figured out yet how to reproduce this wrong behaviour reliably, but it is provoked most often when the gate is turned on and then off in quick succession.

[Save state](https://github.com/awwbees/BespokeSynth/files/7188105/notegate-fail.bsk.zip) with the patch shown above.

NoxyNixie:
Or well I think it resolves it.

In my testing I was able to reproduce the note hanging before I applied this patch and after this patch I couldn't. I'm not sure if this is super scientific since the reproduction isn't super easy to trigger. But I tried significantly longer after the patch with no luck (or well good luck since it didn't happen).

Resolves: #163. 

Edit: Bleh I accidentally added another PR Fixed it though :P

### DIFF
--- a/Source/GridModule.cpp
+++ b/Source/GridModule.cpp
@@ -314,7 +314,11 @@ void GridModule::PostRepatch(PatchCableSource* cableSource, bool fromUserClick)
    if (cableSource == mGridOutputCable)
    {
       auto* target = dynamic_cast<GridControlTarget*>(cableSource->GetTarget());
-      if (target)
+      if (target == mGridControlTarget)
+      {
+         cableSource->Clear();
+      }
+      else if (target)
          target->SetGridController(this);
    }
 }
--- a/Source/GridModule.cpp
+++ b/Source/GridModule.cpp
@@ -314,11 +314,7 @@ void GridModule::PostRepatch(PatchCableSource* cableSource, bool fromUserClick)
    if (cableSource == mGridOutputCable)
    {
       auto* target = dynamic_cast<GridControlTarget*>(cableSource->GetTarget());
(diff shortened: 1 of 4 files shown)


awwbees:
this isn't a correct fix in all scenarios, ignoring the input `time` parameter and calling `NextBufferTime()` is the red flag. I'll need to take a look at this one to find another fix. I've also got a future voice management system planned that should resolve stuck note issues like this more cleanly.

NoxyNixie:
Ah yeah I see, would adding an optional additional `time` argument to the `NextBufferTime` be an appropriate intermediate solution?

Like done in the latest commit to this branch?

awwbees:
no, this overcomplicates the `NextBufferTime()` interface

NoxyNixie:
Okay one more attempt ;)

If this isn't what you want either feel free to reject this.

awwbees:
there's a bit of a bummer here where the note will be played a buffer late, but it's preferable to a stuck note. when I create the new voice system, I should be able to clean this up. thanks for the fix!

Issue #228, opened 2021-09-22
noterangefilter bug


Korblys:
minor issue regarding noterangefilter, unsure if user error or problem with effect itself- when selecting between values 58 and 60 or similar number difference, numbers like 59 aren't selectable via the slider. Attempted to try zoomed in as close as Bespoke allows, yet the same problem persists.

included a recording of issue, showing problem with 58-60 and 62-64.

https://user-images.githubusercontent.com/90932975/134275839-5084ee2c-5d1d-4b65-80c5-2085d0a0daf3.mp4

awwbees:
I will take a look at why it is not giving you full resolution. in the meantime, two options:
- while hovering over the slider, you could use the up and down arrows on the keyboard to increment the value
- while hovering over the slider, you could type the desired number and then press enter

ThePython10110:
You can also hold shift while dragging to change by smaller amounts.

awwbees:
resolved by #1091

PR #1091, opened 2023-01-24
Improved IntSlider and FloatSlider accuracy especially when zoomed in. Fixes: #228.

NoxyNixie:
The conversion to int type was too soon in the process thus reducing accuracy significantly especially when zoomed in. Simply using float's instead of int's in `SetValueForMouse` makes it so much more accurate.


awwbees:
nice find!

PR #1097, opened 2023-01-25
Fix a bug where Canvas Controls could be modulated. Resolves: #951.

### LINKED ISSUE #951: Modulation added to control of canvas based modules crash Bespoke on patch load
## Describe the bug

The bug is applicable to modules having an internal canvas like `notecanvas` and `samplecanvas`.

It is currently possible to link a modulator (`controlsequencer`, `lfo`) to controls of elements of the canvas.
In case of a `notecanvas`, we can click on a note to display its controls. It is then possible to attach a `controlsequencer` to one of the control (lenght, velocity, ...).

It works, the UI will refer to the connection as: "notecanvas\~canvas_controls\~velocity" 
Note: there is no info in the path about the canvas element, how will it know which note is concerned?

If a patch is saved with this setup, it is no longer possible to open it because Bespoke crash with the following error:
> error: Couldn't find UI control at path "notecanvas\~canvas_controls\~velocity"

The error is only visible with a debug build.
Note: the behavior is not exactly the same with an `lfo`, Bespoke will not crash but the link is not restored.

There are others bug reports regarding crashes with corrupted `bsk` (#203, #312, #748), I was able to reproduce theses cases with the provided bsk but in these cases the error is different:
> error: Couldn't find module at path "effectchain\~delay2"
> terminate called after throwing an instance of 'UnknownModuleException'

## Build info

I'm able reproduce the issue with the latest arch package: 1.1.0 (Sep 15 2022 20:52:56)
And also with a fresh build from git: DEBUG BUILD 1.1.0 (Nov 4 2022 11:31:04) (git hash: 4c00c8b1)

## Steps To Reproduce

- Create a new patch
- Add a `notecanvas`
- Add a note on the canvas
- Click the note
- Add a `controlsequencer`
- Link the `controlsequencer` to one of the note's control (like velocity)
- Save the patch
- Load the patch

With the steps above Bespoke crash.
As I said above the crash occur  with a `controlsequencer`. With an `lfo`, the patch can be opened but the link is not restored.

I'm expect Bespoke to be able to open the patch and restore the link between the modulator and the control.
Or at least, it should not be possible to create these kind of links because it corrupts the `bsk` and it is difficult to guess why without a debug build.

NoxyNixie:
Fix a bug where Canvas Controls could be modulated (these controls point to different things in the canvas depending on what is selected). Resolves: #951.

### DIFF
--- a/Source/SongBuilder.cpp
+++ b/Source/SongBuilder.cpp
@@ -104,6 +104,7 @@ void SongBuilder::CreateUIControls()
    for (int i = 0; i < kMaxSequencerScenes; ++i)
    {
       DROPDOWN(mSequencerSceneSelector[i], ("scene" + ofToString(i)).c_str(), &mSequencerSceneId[i], 80);
+      mSequencerSceneSelector[i]->SetNoCableTarget(true);
       mSequencerSceneSelector[i]->SetDrawTriangle(false);
       UIBLOCK_SHIFTRIGHT();
       TEXTENTRY_NUM(mSequencerStepLengthEntry[i], ("bars" + ofToString(i)).c_str(), 3, &mSequencerStepLength[i], 1, 999);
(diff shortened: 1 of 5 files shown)


Source/SongBuilder.cpp
```diff
@@ -985,10 +985,10 @@ void SongBuilder::ControlTarget::CreateUIControls(SongBuilder* owner)
    mCycleDisplayTypeButton = new ClickButton(owner, "type", -1, -1);
    mColorSelector = new DropdownList(owner, ("color" + ofToString(mId)).c_str(), -1, -1, &mColorIndex, 25);
 
-   // Block modulation calbes from connecting to these controls as it behaves wrong (and saves incorrectly) except for the color button, that one is fun and works.
-   mMoveLeftButton->SetNoCableTarget(true);
-   mMoveRightButton->SetNoCableTarget(true);
-   mCycleDisplayTypeButton->SetNoCableTarget(true);
+   // Block modulation cab;es from connecting to these controls as it behaves wrong (and saves incorrectly) except for the color button, that one is fun and works.
```
NoxyNixie, line 988: Hah, this can happen :P

PR #1098, opened 2023-01-25
Changed the behavior or DropdownList's so that they close when you click them again. Resolves: #649.

### LINKED ISSUE #649: Clicking on dropdown opens it, but clicking again does not close it
I am on macOS Big Sur, Bespoke v1.1 and when I click on a dropdown menu it opens, however when I click on it again it does not close. The only way to close it is to click on empty space, which is not the commonly expected UI behavior for dropdowns generally.

NoxyNixie:
Changed the behavior or DropdownList's so that they close when you click them again. Resolves: #649.

I'm not entirely sure why the drop down lists in the titlebar menu don't behave the same so I have a sneaking suspicion my solution isn't the right one even though it works for everything else (that I've tested).

### DIFF
--- a/Source/DropdownList.cpp
+++ b/Source/DropdownList.cpp
@@ -367,6 +367,13 @@ void DropdownList::OnClicked(float x, float y, bool right)
    if (mElements.empty())
       return;
 
+   auto a = TheSynth->GetTopModalFocusItem();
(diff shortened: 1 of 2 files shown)


awwbees:
alternate fix in #1144

PR #1106, opened 2023-01-26
Added a little distinguishable additional graphic to the module receiving pips.

NoxyNixie:
At 2x zoom:
image

At 1x zoom:
image

At 8x zoom:
image

I was trying to strike a balance between how distracting and how distinguishable they are and I hope I did it but well, this is all opinion based and people will like different things. I do know that this should for the most part eliminate confusion between note and pulse receiver pips and help the color blind folks a little.

### DIFF
--- a/Source/IDrawableModule.cpp
+++ b/Source/IDrawableModule.cpp
@@ -373,24 +373,29 @@ void IDrawableModule::Render()
    ofPushStyle();
    const float kPipWidth = 8;
    const float kPipSpacing = 2;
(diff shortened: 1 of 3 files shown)


awwbees:
I find that these changes make the input hints look like too much of a "thing" unfortunately. it draws too much attention to them and makes it look like you're supposed to directly interact with them. maybe instead I'll add an "accepts: [types]" line to the module tooltip, like can be seen on the docs site:
image

PR #1163, opened 2023-02-15
Fixed a bug where note off's would turn off all notes with the same pitch even when specific voice id's were used in modules that use the PolyphonyMgr.

NoxyNixie:


NoxyNixie:
Gah, spotted some mistakes and I think I want to do it slightly differently but have to head to bed now.

Will likely fix this tomorrow.

awwbees:
I have a new voice management system planned for the future that will clean up a whole bunch of issues like this, so I probably am not going to merge this change.

NoxyNixie:
I would in this case still merge this fix as this will fix a bug now as opposed to some unknown time in the future.

Now if you have the rewrite in a hidden branch somewhere and it is nearly finished then sure this fix might be obsolete soon. But as it stands right now it is a fix that will make Bespoke less buggy. Also this fix would make it into 1.2 while the new voice system might not (only you know ;) )?

awwbees:
ok, yeah, this seems like a good fix, which will map nicely to the future voice management work (that work won't be started until after 1.2.0). having two different `Stop()` interfaces had set off my spidey-senses, but it looks like the one without a voice parameter isn't necessary, so removing that makes this change feel much clearer.

it does seem like there's a nonzero chance that this change could result in new stuck note bugs (since `Stop()` is now more selective about which voices get stopped), so that'll be something to keep an eye on

PR #1167, opened 2023-02-17
Remove an additional save buffer halving the memory needed for the `record audio` buffer.

NoxyNixie:
I tested this a bunch and the behavior seems to be mostly identical but with much less memory usage.

I initially thought this buffer was there for threading the actual writing to file but the entire code that does the saving is scoped mutexed so this is not the case.

I think this fix is fine but am not super certain as my knowledge of Bespoke is still limited.

Some before and after memory:

Taskmgr_20230217_173936
Taskmgr_20230216_203251
Taskmgr_20230216_204242

### DIFF
--- a/Source/ModularSynth.cpp
+++ b/Source/ModularSynth.cpp
@@ -96,8 +96,6 @@ ModularSynth::~ModularSynth()
    DeleteAllModules();
 
    delete mGlobalRecordBuffer;
-   delete[] mSaveOutputBuffer[0];
-   delete[] mSaveOutputBuffer[1];
    mAudioPluginFormatManager.reset();
    mKnownPluginList.reset();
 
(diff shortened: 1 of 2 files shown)


awwbees:
this PR doesn't work as-is because the rolling buffer in `gGlobalRecordBuffer` wraps around, so it needs to rearrange the output to save it properly. like this:
n = new data from this trip through the buffer
o = old data from the prior trip through the buffer
t = current time
record buffer: nnnnnnnnnnntooooo
desired output file contents: ooooonnnnnnnnnnn
output file contents from this PR: nnnnnnnnnnnooooo

the removed `GetSample()` loop is the thing that was doing this rearranging.

but, this PR makes the excellent point that the current implementation is incredibly wasteful. I should figure out if there's a way I can do this saving work in-place, rather than keeping a massive buffer allocated just for the saving. (also noteworthy: the reason that massive buffer is always loaded, rather than just allocating right before save, was to avoid the issue of "that jam was great, I'll save it now!" and then crashing because of a memory allocation failure when we're trying to save)

awwbees:
#1207 gets the achieves the memory usage fix while still saving the file in the correct order. thanks for pointing this one out! the global record buffer save process is ancient code (probably 10 years old!) and it wasn't even on my radar as a potential memory sink.

Issue #1160, opened 2023-02-12
Sending an integer to /bespoke/control/vstplugin~preset when preset value is at --- does not change preset value 


ShamanicArts:
Problematic Behaviour:

as stated in Title, VST presets do not change on OSC integer sent unless a preset has already  been manually  selected. this is problematic as it means everytime a template is loaded every single VST in that template has to be manually selected to a preset before they can be controlled externally, this could result in 100s of actions being necessary before a template is fully functional again, in my current template I have at least 30 controls like this. Imagine how tedious this could be in a live performance if the software crashes . 

Expected Behaviour: 

Regardless of what state the preset menu is at at the beginning of template launch, I believe that sending an integer value to the VST should result in a preset change to that preset, rather than being dependent on a preset being loaded into the list for preset switching to be initialised. wether the solution involves changing it so VSTS load with whatever preset they were last on ( or more accurately, whatever state the plugin is in with whatever named preset was in the list loaded in ) or the code is adjusted to allow for preset switching when preset state is --- , I think a feature like this is necessary for Bespoke preset system to be functional.

Issue #1157, opened 2023-02-11
Randomized Order for Presets in VSTs


ShamanicArts:
Currently VST presets do not populate the preset menu in a linear order ( numerical, alphabetical etc ) and instead seem to populate the preset menu at random. this creates a problem when you are trying to map controls of that preset list, either through a value setter or externally via midi or OSC ,  as adding new presets to the list can cause the list order to change randomly meaning that any system would have to be redone on each preset added. 

a better implementation would allow for ordering of the presets based on some criteria, most likely the standard of special characters, numbers then alphabetically. this would allow for consistency in creating presets in the preset menu

ShamanicArts:
Snapshots tied to the plugin module are also impacted by this randomized preset order, a snapshot that captures a plugin state with "A1-preset" loaded, which is the 1st preset in the dropdown in this example, will load a different preset if the order has been affected by this bug , as snapshots pull their data from the index of the list.

PR #1168, opened 2023-02-17
Sort the file list in the vstplugin's preset dropdown.

NoxyNixie:


NoxyNixie:
Tagged another fix onto this PR it fixes being able to modulate the preset dropdown box on vstplugins after load through MIDI/OSC or other modulation without having to click the dropdown.

This PR now resolves #1160 and #1157.

Issue #879, opened 2022-07-18
Midi controller pages not saved correctly


nay-seven:
i have created 2 pages for a Launchpad mini mk1, but when reloading the patch and select the page 0 , grid are not linked anymore

## Build info

bespoke NIGHTLY 1.1.0 (Jul  9 2022 04:38:53) - Mac OSX 10.14.6 - HEAD 2c959c6

## Steps To Reproduce

[see video](https://www.dropbox.com/s/hg5tk6mvfbjs8pg/pages%20bug.mp4?dl=0)

and here the patch+custom json, wip

[bug pages.zip](https://github.com/BespokeSynth/BespokeSynth/files/9130975/bug.pages.zip)

PR #1173, opened 2023-02-19
Add grid connections only after layout is loaded

sbungartz:
This fixes a problem I had, when trying to save/load grid connections on multiple pages of a midi controller (#879 sounds related).
Turns out this is only a problem if a custom layout is used for the controller.

## The issue
When a controller layout other than the one named exactly like the midi device is used in a saved state, connections on pages that were not active when saving are not restored.

This happens because, the grid connections were added in `Init`. At that point, the controller uses the default layout (i.e. the one named like the device or the actual default without any grids). Afterwards the UI State is loaded in `LoadState`, and then the selection from the layout dropdown is restored causing `mGrids` to be (re-)set to the correct setup.

## The fix
Now, we load the grids in `LoadState` after the UI has restored the state of the layout dropdown, so grid connections are always loaded into the correct grids.

I did this very conservatively here, only changing the order for grid connections, because I was not sure if moving the `AddControlConnection` call into `LoadState` for non-grid connections could cause other problems.
This caused some not-so-pretty copy-pasting. If it is safe to just move all `AddControlConnection` calls to `LoadState` that would seem a bit nicer.

### DIFF
--- a/Source/MidiController.cpp
+++ b/Source/MidiController.cpp
@@ -114,11 +114,6 @@ void MidiController::Init()
    mConnections.clear();
 
    mHasCreatedConnectionUIControls = false;
-   // Only add non-grid connections, because the persisted layout has not been loaded yet.
-   // The grid connections will be added in LoadState.
-   for (int i = 0; i < mConnectionsJson.size(); ++i)
-      if (!mConnectionsJson[i].isMember("grid_index"))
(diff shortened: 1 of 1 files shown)


awwbees:
thanks for this change! I need to find some time to dive in and investigate that proposal to move `AddControlConnection()` calls to `LoadState()`, agreed that it would be cleaner to do everything in one spot if possible

PR #1178, opened 2023-02-24
Fix EnvelopeEditor modulation when minimized.

### LINKED ISSUE #1175: Pinned envelopeeditor stop responding to changes when collapsed
## Describe the bug

Pinned envelopeeditor stop responding to changes when collapsed.

## Build info

Linux
Debian Sid
Latest Nightly 84b0c0d

## Steps To Reproduce

- Add an oscillator
- Pin both amp and filter envelopes to show their envelopeeditor
- Connect a macroslider to the decay stage time of both envelopeeditor
- Click to collapse the envelopeeditors
- Change value with the macroslider
- envelopeeditors don't change decay time
- click to uncollapse envelopeeditors, now they start to change decay time when that value is changed with the macroslider

https://user-images.githubusercontent.com/114294020/220308782-fe9680cc-5aae-42cd-a8f8-d5a59a5ac4c8.mp4

envelopeeditor should respond to changes when collapsed, right now it seems to not respond to changes when collapsed

NoxyNixie:
This fix is likely not the correct way to resolve this issue however seems to be necessary at this time since sliders calculate their stuff when rendered.

In theory the correct fix is to remove any and all state changing calculations from rendering code but this is likely far more work than I can already see it being.

Resolves: #1175.

Another "solution" would be to prohibit minimizing of the envelope editor :P

(Also looking at this does this mean in situations like this the actual update rate of the final value is rate limited to the frame rate instead of audio rate?)

### DIFF
--- a/Source/EnvelopeEditor.cpp
+++ b/Source/EnvelopeEditor.cpp
@@ -473,9 +473,6 @@ void EnvelopeEditor::DrawModule()
       ofPopStyle();
    }
 
-   if (Minimized())
-      return;
-
    mMaxSustainSlider->SetShowing(mADSRDisplay != nullptr && mADSRDisplay->GetADSR()->GetHasSustainStage());
 
    mADSRViewLengthSlider->Draw();


awwbees:
> (Also looking at this does this mean in situations like this the actual update rate of the final value is rate limited to the frame rate instead of audio rate?)

ah, yes... since the sliders are on a separate module, that means that `ComputeSliders()` calls won't hit them (which is how sliders' audio-rate updating generally happens), and the only thing causing them to recompute is just the side effect of rendering them. this'll have to be addressed at some point.

PR #1179, opened 2023-02-25
Make file extensions mostly case insensitive on all platforms.

### LINKED ISSUE #732: Samples with upper file name case in extension aren't shown in the load window
## Describe the bug

I just want to use a sample file e.g. sample.WAV, but it isn't shown, have to rename it first to sample.wav.
(Upper case is happening for me when I copy samples from my PCM recorder to a Linux computer.)

## Build info

1.1

couldn't copy build info, as this feature (copying build info) just hangs the application for me

## Steps To Reproduce

- rename a .wav sample extension to .WAV
- try to load it into a sampler

Expected: the .WAV sample in the Bespoke-run file manager
Actual: only .wav samples are shown

NoxyNixie:
- Make the SampleBrowser case insensitive on all platforms regardless of file system case sensitiveness.
- Also accept full capitalization extensions on case sensitive file systems in file chooser dialog's.
- Added *.midi as valid extension for midi files.

Many sources of sample or midi files have the extensions in full capitalization and thus are currently impossible to load directly into bespoke (with the exception of drag and drop) on systems that have a case sensitive file system.

This PR should fix all of the cases where the extension is full capitalization however it won't fix all cases for when the file extension has mixed capitalization (this would require the `juce::FileChooser` to support this which it does not).

I've only done this for FileChooser's that can accept files from other sources than only Bespoke itself (i.e. I'm not doing it for *.pfb or *.vstp since these are made by Bespoke always lowercase).

I have tested some of this code but I can't truly test it since I do not have a case sensitive file system.

Resolves: #732.

### DIFF
--- a/Source/NoteCanvas.cpp
+++ b/Source/NoteCanvas.cpp
@@ -596,9 +596,9 @@ void NoteCanvas::QuantizeNotes()
 void NoteCanvas::LoadMidi()
 {
    using namespace juce;
-   String file_pattern = "*.mid,*.midi";
+   String file_pattern = "*.mid;*.midi";
    if (File::areFileNamesCaseSensitive())
-      file_pattern += file_pattern.toUpperCase();
+      file_pattern += ";" + file_pattern.toUpperCase();
    FileChooser chooser("Load midi", File(ofToDataPath("")), file_pattern, true, false, TheSynth->GetFileChooserParent());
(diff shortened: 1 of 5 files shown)


awwbees:
makes sense to me. I don't have a convenient setup to test case sensitivity on, so let's just commit it and see how it works for people.

PR #1180, opened 2023-02-27
Prefabs (or containers) were saving twice.

NoxyNixie:
- This was exposed by the snapshots module since it was exponentially increasing its memory with references to these duplicate modules.
- Modernized some portions of the Snapshot code.

I'm not sure this is ready for merging yet but I do want this to be up so that comments can be given.

The thing I'm not super sure about is if this change breaks working savestates. So far I've been able to load many savestates fine older and newly created ones but I also know that this testing is not at all inclusive.

There are some savestates that I think should be working but not sure, I do have quite a few savestates that I know are bad.

So this will still need some work, just not sure how/where yet. 

Edit: Owh and looks like I left some debug stuff in there too.

### DIFF
--- a/Source/IDrawableModule.cpp
+++ b/Source/IDrawableModule.cpp
@@ -1129,7 +1129,9 @@ void IDrawableModule::SaveState(FileStreamOut& out)
    }
 
    if (GetContainer())
+   {
       GetContainer()->SaveState(out);
+   }
    else
    {
       out << (int)mChildren.size();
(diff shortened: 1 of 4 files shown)


NoxyNixie:
Alrighty I have tested a bunch more and with my latest commits all my older savestates (that should load fine) load fine.
All the savestates that used to have the bad/incorrect data for snapshots are now also automatically fixed.

So I think this is ready for review.

awwbees:
I know that I'm a dinosaur, but so much of that modernization pass makes the code read less clearly to me. I removed it from this PR to avoid distracting from the functional part of the fix being introduced here.

PR #1189, opened 2023-03-03
Remove cables targeting controls on a deleting module.

### LINKED ISSUE #222: deleting a module with modulator does not delete the modulator cable
pretty easy to reproduce:

- create any module that has a modulatable parameter
- attach a modulator or pin an lfo to it
- delete the module

the wire connecting the modulator to the module will still be there, just floating in space.

NoxyNixie:
Remove cables targeting controls on a deleting module. Resolves: #222.


awwbees:
nice, good fix for a longstanding issue that I kept forgetting to look into

PR #1193, opened 2023-03-06
Allow disabling of output limit.

NoxyNixie:
Also brought some of the code to modern standard and used the same method all across the plugin (std::clamp and CLAMP where both used).

Also applied an optimization done in 806b123a21a0bf1cf9abce7eec14ab01c56f0c63 to other variables used in the tight loops but I have not actually profiled before and after so have no idea if this is better or not or if the compiler already did this on its own. At least it should now be consistent between single channel output modules and double channel output modules.


awwbees:
sorry, the addition of the modernization stuff is distracting me from the real contents of this change

PR #1198, opened 2023-03-10
Added VU meters to audiometer

prehensile:
as discussed on Discord: https://discord.com/channels/846834873254805554/854199300540071956/1083425235698724994

Added optional VU meters to  the `audiometer` module, toggleable with a checkbox. One or two VU meters are displayed, depending on whether the input signal is mono or stereo. Without the toggle enabled, the module has the same pixel dimensions as unaltered `audiometer` module.

Meters are mostly just copypasted from the `output` module. I've moved some consts and function calls out of for loops. 

Tested on Windows.

### DIFF
--- a/Source/AudioMeter.cpp
+++ b/Source/AudioMeter.cpp
@@ -31,12 +31,19 @@ AudioMeter::AudioMeter()
 : IAudioProcessor(gBufferSize)
 {
    mAnalysisBuffer = new float[gBufferSize];
+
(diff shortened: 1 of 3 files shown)


awwbees:
thanks for the idea here, apologies for taking to long to get to this. I ended up creating this functionality in a way that shared the level meter code (rather than a copy-pasted duplication) in #1727

PR #1202, opened 2023-03-12
Trying to fix a few race conditions

neuschaefer:
It seems to be a common problem in Bespoke that there are race conditions between the render thread and other threads that update various variables.

For this pull request, I used ThreadSanitizer and fixed two race conditions, to get familiar with the process, and gather some feedback from other Bespoke developers.

### DIFF
--- a/Source/LocationZoomer.cpp
+++ b/Source/LocationZoomer.cpp
@@ -46,6 +46,8 @@ void LocationZoomer::Init()
 
 void LocationZoomer::Update()
 {
+   auto g = TheSynth->LockRenderWithGuard();
+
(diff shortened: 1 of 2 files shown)


Source/LocationZoomer.cpp
```diff
@@ -46,6 +46,8 @@ void LocationZoomer::Init()
 
 void LocationZoomer::Update()
 {
+   auto g = TheSynth->LockRenderWithGuard();
```
neuschaefer, line 49: This is not good. I have now seen a deadlock between the most threads except for the audio thread, involving this line.

PR #1215, opened 2023-04-03
snapshots: add option to auto-store when switching between snapshots

awwbees:


Source/Snapshots.cpp
```diff
@@ -439,6 +453,15 @@ void Snapshots::Store(int idx)
...
 }
 
+void Snapshots::Delete(int idx)
+{
+   assert(idx >= 0 && idx < mSnapshotCollection.size());
+
+   SnapshotCollection& coll = mSnapshotCollection[idx];
+   coll.mSnapshots.clear();
+   coll.mLabel = ofToString(idx);
+}
```
NoxyNixie, line 464: What about the labels in `mCurrentSnapshotSelector`?

awwbees: good catch!

PR #1254, opened 2023-06-05
Allow the scrollwheel to adjust the Q slider while hovering over a dot in the `eq` module.

NoxyNixie:
You can also adjust the increment by holding shift or control/command for faster and slower respectively.

### DIFF
--- a/Source/EQModule.cpp
+++ b/Source/EQModule.cpp
@@ -394,6 +394,25 @@ bool EQModule::MouseMoved(float x, float y)
    return false;
 }
 
+bool EQModule::MouseScrolled(float x, float y, float scrollX, float scrollY, bool isSmoothScroll, bool isInvertedScroll)
+{
+   if (mHoveredFilterHandleIndex != -1)
+   {
+      float add = scrollY;
+      if (GetKeyModifiers() & kModifier_Command)
+      {
+         add /= 10;
+      }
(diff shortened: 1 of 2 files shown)


PowerUser64:
The only things I would change about this are currently, holding shift while scrolling changes the q in larger increments rather than smaller increments, which is inconsistent with the rest of bespoke. Additionally, adjustments should probably be larger for higher values and lower for small values, right now it's linear. Other than that, things work as expected!

NoxyNixie:
> The only things I would change about this are currently, holding shift while scrolling changes the q in larger increments rather than smaller increments, which is inconsistent with the rest of bespoke. 

Owh looks like I reversed that logic by accident.

> Additionally, adjustments should probably be larger for higher values and lower for small values, right now it's linear. Other than that, things work as expected!

Ah yeah, non linear does feel better. I tweaked some values and personally think it feels good now but what is your opinion?

PowerUser64:
Feels great! If I could change one other thing, it would be to have pressing backslash on a point could reset its values. Other than that, scrolling feels great! Shift-scrolling works as expected, and changing the max/min values produces expected results as well.

NoxyNixie:
The backslash key now reset to original and also added `[` and `]` to behave similar to sliders where they halve and double.

PR #1258, opened 2023-06-17
push2 features and other fixes

awwbees:
-add ability to quickly spawn connected looper or notelooper from push2
-add notelooper push2 interface
-add drumplayer slot editing on push2
-change push2 interface to select output cable
-allow module insertion into chain via push2 by holding shift while patching
-after resetting parameter via push2, add cooldown to not accept change for the next second, to avoid accidentally changing the value you were trying to reset
-remove 0 from ADSR sliders to avoid the ability to cause NaNs
-fix notecanvas arrows causing notes to double-move

### DIFF
--- a/Source/ADSRDisplay.cpp
+++ b/Source/ADSRDisplay.cpp
@@ -49,10 +49,10 @@ ADSRDisplay::ADSRDisplay(IDrawableModule* owner, const char* name, int x, int y,
    if (floatListener)
    {
       int sliderHeight = h / 4;
-      mASlider = new FloatSlider(floatListener, (std::string(name) + "A").c_str(), x, y, w, sliderHeight, &(mAdsr->GetA()), 0, 1000);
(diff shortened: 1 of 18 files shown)


Source/ADSRDisplay.cpp
```diff
@@ -49,10 +49,10 @@ ADSRDisplay::ADSRDisplay(IDrawableModule* owner, const char* name, int x, int y,
    if (floatListener)
    {
       int sliderHeight = h / 4;
-      mASlider = new FloatSlider(floatListener, (std::string(name) + "A").c_str(), x, y, w, sliderHeight, &(mAdsr->GetA()), 0, 1000);
-      mDSlider = new FloatSlider(floatListener, (std::string(name) + "D").c_str(), x, y + sliderHeight, w, sliderHeight, &(mAdsr->GetD()), 0, 1000);
+      mASlider = new FloatSlider(floatListener, (std::string(name) + "A").c_str(), x, y, w, sliderHeight, &(mAdsr->GetA()), 0.001f, 1000);
```
NoxyNixie, line 52: hhmm, I have a feeling this fix is not the correct fix. Shouldn't at the least the attack be capable of being 0 for super snappy things?

PR #1337, opened 2023-07-03
MIDI control for Arturia MiniBrute.

jberentsson:


awwbees:
sorry, there are several aesthetic issues with this change, and likely functional issues that I have no way to test since I don't have a minibrute, so I can't accept this one

image

potential functional issue:
- "on" and "off" seem to be swapped in code for "arpeggiatorHold" and "localOnOff", with on = 0 and off = 1. the likely fix is just to swap these, but I don't have a minibrute, so I can't confirm this

aesthetic issues:
- there are a bunch of radiobuttons with no labels or implication to what the controls do
- the controls are all capitalized, but the rest of bespoke is all lowercase
- inconsistency between "ON/OFF" and "Off/On" (both capitalization inconsistency, and ordering inconsistency)
- inconsistency between abbreviating "Lin" "Log" vs spelling out "Linear" and "Logarithmic"
- member variables don't match bespoke code style (should be `mMemberVariable`)

my inability to test the changes with the hardware means I'd just have to accept this PR on blind faith, but these various issues make me question the care put into this change, so I'll have to turn it down. sorry!

PR #1347, opened 2023-07-08
Initialize mod wheel and pressure as 0 instead of 0.5. Resolves: #1344.

### LINKED ISSUE #1344: Notelooper and Drumplayer modules not working as expected
## Describe the bug

On freshly opened project Drumplayer plays weird weak, squeezed sounds when Notelooper sends pattern to it.

## Build info
bespoke NIGHTLY 1.1.0 (Jul  5 2023 14:14:30) - Linux - HEAD 4de7652

## Steps To Reproduce

In my setup I plug Drumsequencer -> Notelooper -> Drumplayer modules. When I play pattern on Drumsequencer - Drumplayer sounds Ok. When record this pattern to some slot on Notelooper - Drumplayer sounds Ok. Turn off Drumsequencer, only Notelooper plays recorded pattern, Drumplayer sounds Ok. Save. Reopen project. Notelooper plays pattern, and Drumplayer plays weird. Squeezed, cut, weak sounds. Then I clean Notelooper, and redo everything from beginning, Sounds Ok again until I save and open next time. Every time when Notelooper plays prerecorded pattern on freshly opened project - Drumplayer sounds weak and squeezed.

It should play normal sounds :)

NoxyNixie:
Initialize mod wheel and pressure as 0 instead of 0.5. Resolves: #1344.

It seems that 0.5 was chosen as being in the middle of the mod wheel but I think most of the code assumes 0 is the middle. I'm not sure this also applies to the pressure thing.

The actual issue is likely that modulation isn't properly saved/loaded in CanvasElements but I don't even know where to start looking for that (the saving and loading of modulation seems to be not there at all? or I'm missing where it is supposed to be?).


awwbees:
503b8c255c92cc418f91fb822a21657312230a26 fixed a different issue, and accepting this PR would break that. #1344 will require a different fix.

PR #1352, opened 2023-07-09
Add a setting that allows changing the behavior of pasting onto UI controls.

NoxyNixie:
Add a setting that allows changing the behavior of pasting onto UI controls.

I left it default off to preserve the existing behavior as default but I find the immediate paste mode far more natural.


awwbees:
thanks! cool feature

PR #1397, opened 2023-07-31
Allow to devendor abletonlink on systems that provide it system-wide

dvzrv:
Most Linux distributions package abletonlink and therefore do not want to rely on a vendored version of the library.
Add the `BESPOKE_SYSTEM_ABLETONLINK` cmake option to allow using a system-wide installed abletonlink.


dvzrv:
I'm using this on Arch Linux for the distribution package and it works as intended.

merkys:
I can confirm this works as intended on Debian as well.

merkys:
Thanks for merging this @NoxyNixie!

PR #1401, opened 2023-08-01
Add a `module.set_targets(targetPathList)` function to Python scripting

tupto:
Adds a new Python scripting capabilty, allowing users to set multiple targets for a given module. Targets are selected using comma separated target paths.

Example:
```python
import module

gate = module.create("notegate", 300, 300)

osc0 = module.create("oscillator", 300, 500)
osc0.set_name("osc0")

osc1 = module.create("oscillator", 600, 500)
osc1.set_name("osc1")

gate.set_targets("osc0,osc1")
```

Attached are some test `.bsk` files, simply run the scripts in each file and the results should be clear. Most notably among these is the `audio routing.bsk` file which shows this change "allows" a module with an audio output to route to multiple modules but only the last to be connected functions. Further details can be found in the comments of that file. If it is deemed necessary, additional development could potentially handle this by creating a new `audiorouter` instance and routing through that.

Big-ups to Noxy for holding my hand through developing this addition.

[set_targets.zip](https://github.com/BespokeSynth/BespokeSynth/files/12234354/set_targets.zip)

### DIFF
--- a/Source/IDrawableModule.cpp
+++ b/Source/IDrawableModule.cpp
@@ -576,11 +576,19 @@ void IDrawableModule::RemovePatchCableSource(PatchCableSource* source)
 
 void IDrawableModule::ClearAllPatchCableSources()
 {
-   if (mMainPatchCableSource != nullptr)
-      mMainPatchCableSource->Clear();
-   else if (!mPatchCableSources.empty())
-      for (auto source : mPatchCableSources)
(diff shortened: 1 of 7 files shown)


NoxyNixie:
> Big-ups to Noxy for holding my hand through developing this addition.

You're welcome ;)

PR #1413, opened 2023-08-20
Allow toggling enabling state of prefabs, this will attempt to toggle the enabled state of all modules contained within.

NoxyNixie:
BespokeSynth_20230820_061056

### DIFF
--- a/Source/Prefab.cpp
+++ b/Source/Prefab.cpp
@@ -26,6 +26,8 @@
 */
 
 #include "Prefab.h"
+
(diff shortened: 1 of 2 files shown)


awwbees:
I'd like to eventually get something in that doesn't directly use `mEnabled`, and is instead a separate layer that preserves the `mEnabled` state... but that'll touch a lot of files and possibly cause bugs. so this seems like a good short-term functionality gain. thanks!

PR #1416, opened 2023-08-24
Added MIDI controller layouts for Akai MPK Mini

denilsonsa:
This should work on both Akai MPK Mini and Akai MPK Mini Play, as both are mostly identical regarding their MIDI capabilities, but the later has a built-in synthesizer. I only have the Mini Play, I'm assuming everything else works similarly for the non-play version. Given how similar these devices are, it's a safe assumption.

https://www.akaipro.com/mpk-mini-mk3
https://www.akaipro.com/mpk-mini-play

I've included the full 0-120 range of MIDI notes for the main piano keyboard. That's because these devices have the `Oct-` and `Oct+` buttons to change which range of notes their keys is sending. Thus, even though it only has 25 keys (2 octaves), it can send the full range of notes. I've organized them in groups of two octaves per row. Thus, the middle row corresponds to (almost) all keys from the keyboard in its default octave, and each two presses of the `Oct-`/`Oct+` will shift the keys one row upwards or downwards.

Both devices only have 8 pads, but both have a button to let those drum pads send up to 16 values. Additionally, even though the MPK Mini Play only has 4 knobs, it has a button toggle to let them behave like 8 knobs. (Sidenote: the knobs are numbered top-down, while the pads are numbered bottom-up, just like the devices themselves.)

The joystick behaves weirdly:

* When "internal sounds" is off it behaves as pitchbend in the horizontal axis and as CC1 in the vertical axis (both directions increasing the value from zero to one).
* When "internal sounds" is on (only possible in the Mini Play), then the pitchbend axis is now the vertical axis, and the horizontal axis reports as CC11 (expression) to the left and CC1 (modwheel) to the right.

Finally, these devices have 8 "Favorite" presets, changing their behavior. Some presets may report different notes for the pads and different codes for the knobs.

For the pads:

* Favorite 4 FL Studio20 remaps the pads to 37, 36, 42, 54, 40, 38, 46, 44, 48, 47, 45, 43, 49, 55, 51, 53; which look somewhat random/arbitrary.
* Favorite 7: MajorPads remaps the pads to 48, 50, 52, 53, 55, 57, 59, 60, 62, 64, 65, 67, 69, 71, 72, 74; which correspond to the white keys of a piano.
* Favorite 8: MinorPads remaps the pads to 45, 47, 48, 50, 52, 53, 55, 57, 59, 60, 62, 64, 65, 67, 69, 71; ; which also correspond to the white keys of a piano.
* All others report the pads exactly as in the JSON of this commit.

For the knobs:

* Favorite 2: AbletonLive remaps the knobs to 1 to 8 (which also clashes with the CC1 for the joystick).
* All others report the knobs exactly as in the JSON of this commit.

Users can also use their proprietary software to customize each preset to their own needs. I don't have access to that proprietary software (requires registering an account with them, then downloading a downloader, then downloading their software — too much hassle). If you're interested in pursing that, here are some open-source projects for similar devices:

* https://github.com/gljubojevic/akai-mpk-mini-editor
* https://github.com/BastiTee/akai-mpkmini-mkii-control
* https://github.com/tsmetana/mpk3-settings

---

If I could, I would include the pads into the main controller layout. Unfortunately, each layout is restricted to a single channel (as far as I understand).

Screenshot of the new layouts


denilsonsa:
Added a new layout for when "Internal Sounds" is enabled:

image

Unfortunately, the two knobs labeled *EQ Low* and *EQ High* map to the same code.

tomc98:
I just want to say thank you! This was super useful I really wish it was merged into main. Works great with my MPK mini

denilsonsa:
> Unfortunately, the two knobs labeled _EQ Low_ and _EQ High_ map to the same code.

Now that I understand more about MIDI, they map to the same code because they are actually sending a NRPN (Non-Registered Parameter Number) value. Which means it sends this sequence of CC messages:

1. CC 99 for the MSB (most significant bits) of the NRPN.
    * Value 55 is sent by both knobs.
1. CC 98 for the LSB (least significant bits) of the NRPN.
    * Value 0 is sent by the EQ Low knob.
    * Value 3 is sent by the EQ High knob.
1. CC 6 for MSB Data Entry. The value in range 0-127 is sent to the NRPN choosen by the previous two steps combined.
1. (not send by this device) CC 38 for LSB Data Entry.

You can see that yourself on this page: https://studiocode.dev/midi-monitor/

It was also documented in this blog <https://sandsoftwaresound.net/akai-mpk-mini-play-midi-implementation/>, where the author studied the chip that generates sounds from the device, and even hacked it by adding some other wires to control that chip directly.

PR #1418, opened 2023-08-25
Bypass rework

NoxyNixie:
Combined commit message:
```
Changed the following modules to use a bypass mechanism when disabled: `gain`, `audiometer`, `fftvocoder`, `dcoffset`, `feedback`, `inverter`, `lissajous`, `looper`, `looperrecorder`, `multitapdelay`, `notetable`, `samplecapturer`, `samplergrid`, `seaofgrain`, `send`, `spectrum`, `signalclamp`, `vocoder`, `waveformviewer`, `waveshaper`, `pulsechance`, `pulsedelayer` and `pulseflag`.
Made `spectrum` and `waveformviewer` function when the target is not connected.
Removed the enable/bypass checkbox on the `splitter`.
Minor rendering optimization when the modules `lissajous`, `spectrum` and `waveformviewer` are disabled/bypassed.
Fixed a bug in the `ringmodulator` module where it would not compute the sliders for every sample correctly.
Fixed a bug with the visualization of the output cable on a `looper`.
Fix a bug that inhibited removing the voicelimit in the `oscillator`, `fmsynth` and `karplusstrong` modules.
Added voiceId to the `notedisplayer`.
Added "pan" to `modulationvisualizer` and made the module resizable.
```

Please don't hesitate to poke me on discord if you have questions or want me to do things differently ;)

### DIFF
--- a/Source/Amplifier.cpp
+++ b/Source/Amplifier.cpp
@@ -46,27 +46,30 @@ void Amplifier::Process(double time)
 {
    PROFILER(Amplifier);
 
-   if (!mEnabled)
+   IAudioReceiver* target = GetTarget();
+
+   if (target == nullptr)
       return;
 
    SyncBuffers();
    int bufferSize = GetBuffer()->BufferSize();
 
-   IAudioReceiver* target = GetTarget();
-   if (target)
+   ChannelBuffer* out = target->GetBuffer();
+   for (int ch = 0; ch < GetBuffer()->NumActiveChannels(); ++ch)
    {
-      ChannelBuffer* out = target->GetBuffer();
-      for (int ch = 0; ch < GetBuffer()->NumActiveChannels(); ++ch)
+      auto getBufferChannelCh = GetBuffer()->GetChannel(ch);
+      if (mEnabled)
       {
-         auto getBufferChannelCh = GetBuffer()->GetChannel(ch);
          for (int i = 0; i < bufferSize; ++i)
          {
             ComputeSliders(i);
             gWorkBuffer[i] = getBufferChannelCh[i] * mGain;
          }
          Add(out->GetChannel(ch), gWorkBuffer, GetBuffer()->BufferSize());
-         GetVizBuffer()->WriteChunk(gWorkBuffer, GetBuffer()->BufferSize(), ch);
       }
+      else
+         Add(out->GetChannel(ch), getBufferChannelCh, GetBuffer()->BufferSize());
+      GetVizBuffer()->WriteChunk(gWorkBuffer, GetBuffer()->BufferSize(), ch);
    }
 
    GetBuffer()->Reset();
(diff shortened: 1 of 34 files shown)


awwbees:
seems great, gonna hold off on this one until after 1.2.1 though since it's pretty significant

NoxyNixie:
Understood.

PowerUser64:
1.2.1 has released, should we revisit this?

For the record, I've been using this for a while and it's great so far. I think the only thing missing is an `audiogate` module, to replace the disable button that used to do this. Something like that can bring the volume to zero quickly without clicks or pops. I probably should make a feature request for this, but it seems fitting to mention it here.

NoxyNixie:
>  I think the only thing missing is an `audiogate` module, to replace the disable button that used to do this. Something like that can bring the volume to zero quickly without clicks or pops.

I did not see the need for this module since you can do this with the `gain` module. Add a very small smooth value (hold alt while adjusting the slider) to the slider and then set it to `0` for off and `1` for on.

PowerUser64:
I see what you mean, but this is also true for many other modules too. For example, the notegate can be achieved in the same way by setting a notechance set to zero. I think it's still important the notegate exists because it's a nice abstraction that makes sense to exist. Even though an audiogate could be created from modules through any number of methods, I still believe it makes sense for it to exist as its own module.

PowerUser64:
I have been using this for some time now, and the only issue that I notice is under certain circumstances, a disabled module's cable will be rendered as flat even when signal is going through it. It's a bit tough to reproduce, but the main module I notice it with is the `gain` module, not sure if it applies to others as well. I'll see if I can get a screenshot next time I notice it.

PowerUser64:
Here's a way to get something similar to what I just described, possibly related.

image

NoxyNixie:
image

I think I fixed the visualization of cables after a gain module when bypassed. @PowerUser64 have you seen any other modules where this happened?

PowerUser64:
I believe I have seen a few others, but I can't remember which ones off the top of my head (it'd take some investigating that I might be able to do at some point). I think the waveformviewer miiight have been one of them? I could be wrong though (and I can't test it right now).

NoxyNixie:
hhmm I can't reproduce with the waveformviewer:

image

Any other modules?

NoxyNixie:
I tested all the modules that this PR is touching and I think the incorrect cable rendering was only present in the `gain` module which I fixed with [2f72092](https://github.com/BespokeSynth/BespokeSynth/pull/1418/commits/2f7d09217903989687df74f2ec1622150d020f26).

So I think this should be ready to merge in.

PowerUser64:
Nice! I must have done something else weird to get the waveformviewer to do whatever I was thinking of. I think this is good to merge!

Oh, and by the way I discovered there is a module to block the signal flow, it's the `muter`! It's a bit large (an effectchain module), but it gets the job done!

image

awwbees:
sorry for the (very significant) delay in merging this, thanks for cleaning all of this up!

NoxyNixie:
All good, late is better than never.

PR #1420, opened 2023-08-27
Add `audiosplitter` module.

NoxyNixie:
Also made the splitter modules non minimizable. It seems off to me to allow these to be minimized. The size difference is negligible and the cables become harder to follow.


Source/AudioSplitter.h
```diff
@@ -0,0 +1,63 @@
...
+   static bool AcceptsNotes() { return false; }
+   static bool AcceptsPulses() { return false; }
+
+   void CreateUIControls() override;
+   bool CanMinimize() override { return false; }
+
+   //IAudioSource
+   void Process(double time) override;
+   void PostRepatch(PatchCableSource* cableSource, bool fromUserClick);
+   int GetNumTargets() override { return 0; }
```
awwbees: this line seems unusual, needs some investigation to understand if it'll cause any subtle issues

NoxyNixie: hhmm, the only thing I haven't tested where I've seen this used is the circular dependency checker.

NoxyNixie:
The circular dependency checker now works with the `audiosplitter` module.

Issue #1340, opened 2023-07-05
Bug - granulator grains click when being deleted


PowerUser64:
## Describe the bug

When grains in the granulator stop playing sound, they create clicking noises because they are not faded out.

## Build info

Click the `?` in the upper right corner of Bespoke, click "copy build info", then paste that here.

## Steps To Reproduce

1. Make something like this:
![image](https://github.com/BespokeSynth/BespokeSynth/assets/24578572/c353784b-7650-48f6-a7d2-920a964385de)
2. This should make no sound but it does

## Expected vs. actual behavior

The granulator should have a fade control to allow for a small fade in/out to be added to the start and end of each grain or fade in each grain automatically by automating its volume with a slew limiter (or such) so they don't create clicks. Right now, each grain creates clicking noises which can be disruptive to using it for ambient sound.

PowerUser64:
Fixed in #1427!

PR #1427, opened 2023-09-05
Fixed a bug in the cosine window of the granulators that caused clicky noises at the end of grains.

NoxyNixie:
We figured out the approximation of the cosine function was off at the tailing end of the grains and found the following in the JUCE documentation:

>Provides a fast approximation of the function cos(x) using a Pade approximant continued fraction, calculated sample by sample.
> 
> Note: This is an approximation which works on a limited range. You are advised to use input values only between -pi and +pi for limiting the error.

Since the original code goes to `TWO_PI` we adjusted it to not go that high thus resolving the major inaccuracy of the approximation.


NoxyNixie:
I believe this resolves #1340.

awwbees:
awesome find!

PR #1449, opened 2023-10-24
use temp file to make sure save files don't get corrupted if a crash ever happens mid-save

awwbees:


Source/ModularSynth.cpp
```diff
@@ -1122,8 +1122,8 @@ void ModularSynth::KeyPressed(int key, bool isRepeat)
    mModuleContainer.KeyPressed(key, isRepeat);
    mUILayerModuleContainer.KeyPressed(key, isRepeat);
 
-   if (key == '/' && !isRepeat)
-      ofToggleFullscreen();
+   //if (key == '/' && !isRepeat)
+   //   ofToggleFullscreen();
```
NoxyNixie, line 1126: Was this intended?

awwbees: yes, I was having issues on macOS where it wasn't possible to exit from fullscreen after entering it. presumably some bug introduced by a new macOS version.

Issue #1084, opened 2023-01-24
Bug - can't add modulation to presets when many modules are connected


PowerUser64:
## Describe the bug

The cable port dots cover up the purple dot, making it impossible to add modulation sources to the presets module if there are a lot of modules connected to it

## Build info

commit b0e179b5a6d5c6433965d92a943bb1019187dc15

## Steps To Reproduce

1. Add at least three modules
2. Add a presets module and connect its grey cable to each of the modules
3. Try to use the purple cable to connect the presets module to some parameters (you can't grab it)

![image](https://user-images.githubusercontent.com/24578572/214185892-5268e839-2834-4c78-be16-3fcd9c183ee2.png)

## Expected vs. actual behavior

The purple modulation cable should be grabbable regardless of how many things are attached to the grey cable.

PowerUser64:
resolved by #1455

PR #1455, opened 2023-10-24
fix issue with snapshot cable sources overlapping each other when multiple modules are connected

awwbees:


PowerUser64:
fixes #1084

PR #636, opened 2021-11-21
Allow devendoring of jsoncpp

dvzrv:
CMakeLists.txt:
Add the `BESPOKE_SYSTEM_JSONCPP` option (defaults to `OFF`) to indicate,
that a system version of jsoncpp should be used instead of the bundled
version.

Source/CMakeLists.txt:
Use a conditional generator expression to either link against
`bespoke::json` or against `jsoncpp` (depending on
`BESPOKE_SYSTEM_JSONCPP`).

libs/CMakeLists.txt:
Depending on `BESPOKE_SYSTEM_JSONCPP` include the bundled version of
jsoncpp or use the pkgconfig integration to find a system-wide jsoncpp.

Relates to #590

PR #1459, opened 2023-10-26
Improve location of system-provided jsonpp

merkys:
On some operating systems, in particular Debian and Ubuntu, jsonpp headers do not reside in default header locations. Thus Debian builds with `BESPOKE_SYSTEM_JSONCPP` fail, as jsonpp headers reside under `/usr/include/jsoncpp/` while default header location is `/usr/include/`. A solution would be to ask CMake about jsonpp, which is implemented in this PR. I believe this should improve BespokeSynth's compatibility with operating systems. However, it would be nice to test that on other systems as well. I would be grateful if @dvzrv and @baconpaul could review it as they did with #636.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,6 +81,10 @@ message(STATUS "Python version: ${Python_VERSION}")
 message(STATUS "Python executable: ${Python_EXECUTABLE}")
 message(STATUS "Python library: ${Python_LIBRARIES}")
 
+if(BESPOKE_SYSTEM_JSONCPP)
+    find_package(jsoncpp REQUIRED)
+endif()
+
 if(NOT BESPOKE_SYSTEM_JUCE)
   message(STATUS "Using JUCE from ${BESPOKE_JUCE_LOCATION}")
   add_subdirectory(${BESPOKE_JUCE_LOCATION} ${CMAKE_BINARY_DIR}/JUCE EXCLUDE_FROM_ALL)
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -824,7 +824,7 @@ endif ()
 target_link_libraries(BespokeSynth PRIVATE
     bespoke::exprtk
     bespoke::freeverb
-    $<IF:$<STREQUAL:${BESPOKE_SYSTEM_JSONCPP},OFF>,bespoke::json,jsoncpp>
+    $<IF:$<STREQUAL:${BESPOKE_SYSTEM_JSONCPP},OFF>,bespoke::json,jsoncpp_lib>
     bespoke::leathers
     bespoke::nanovg
     bespoke::psmove


baconpaul:
Could find package check the not found variable and display an error or warning in this case?

I am not sure how to test this. But the change matches your description!

merkys:
> Could find package check the not found variable and display an error or warning in this case?

Fair point, I will try building without jsoncpp on my system and see what happens.

> I am not sure how to test this. But the change matches your description!

One may supply `BESPOKE_SYSTEM_JSONCPP=ON` for CMake build and check if it manages to find system-provided jsoncpp. If it worked before, it should in principle work now, but as said, I can only test on Debian and Ubuntu OSes.

merkys:
> Could find package check the not found variable and display an error or warning in this case?

If jsonpp is not installed, `find_package` fails with the following message:

```
CMake Error at CMakeLists.txt:86 (find_package):
  By not providing "Findjsoncpp.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "jsoncpp", but
  CMake did not find one.

  Could not find a package configuration file provided by "jsoncpp" with any
  of the following names:

    jsoncppConfig.cmake
    jsoncpp-config.cmake

  Add the installation prefix of "jsoncpp" to CMAKE_PREFIX_PATH or set
  "jsoncpp_DIR" to a directory containing one of the above files.  If
  "jsoncpp" provides a separate development package or SDK, be sure it has
  been installed.
```

merkys:
Thanks for merge, @awwbees!

KottV:
It doesn't work on systems without `jsoncppConfig.cmake`. I see this error on Arch and openSUSE.
And there is already pkg-config check for jsoncpp in https://github.com/BespokeSynth/BespokeSynth/blob/3c519d1f9a305866aa1cabd5e72c8541a7c82d20/libs/CMakeLists.txt#L8 which does the job.

dvzrv:
> It doesn't work on systems without `jsoncppConfig.cmake`. I see this error on Arch and openSUSE. And there is already pkg-config check for jsoncpp in
> 
> https://github.com/BespokeSynth/BespokeSynth/blob/3c519d1f9a305866aa1cabd5e72c8541a7c82d20/libs/CMakeLists.txt#L8
> which does the job.

Hi! Sorry, but I didn't have enough time the past weeks to look at this.

I agree with what @KottV writes here and wonder why this change was necessary in the first place.
I'm sure that Debian-based packaging also relies on pkgconfig integration and as such the relevant data can be gotten from there as well.
As is, this breaks all downstreams, that build jsoncpp with meson and not with cmake.

So since cmake has integration for pkgconfig, this is what should be used here instead IMHO.

awwbees:
so should I revert this change? I don't have enough context to understand the intricacies of the issue, so I'll let the participants in this thread decide.

KottV:
I think I have a proper fix for that.

KottV:
Ok, old approach worked for openSUSE and Arch by lucky, just because `json` subdir was in the system `/usr/include` dir, if this dir was in the `/usr/include/jsoncpp_or_whatever` then it would fall as in Debian, because this include path is required from `Source` sourcetree subdir, not from `lib`.
I moved that pkg-config check to the root `CMakeLists.txt` now it seems to work, tested in Arch and Debian.

https://github.com/BespokeSynth/BespokeSynth/pull/1525

merkys:
> Ok, old approach worked for openSUSE and Arch by lucky, just because `json` subdir was in the system `/usr/include` dir, if this dir was in the `/usr/include/jsoncpp_or_whatever` then it would fall as in Debian, because this include path is required from `Source` sourcetree subdir, not from `lib`.

Exactly, this was the reason I made this PR.

> I moved that pkg-config check to the root `CMakeLists.txt` now it seems to work, tested in Arch and Debian.

Thanks a lot, the fix in #1525 seems appropriate.

PR #1469, opened 2023-11-17
Fixed typo in help about the minimap

maxigaz:


NoxyNixie:
Hah, I think this PR is wrong :P we should rename all occurrence of minimap to mimimap it sounds far more fun!

maxigaz:
> Hah, I think this PR is wrong :P we should rename all occurrence of minimap to mimimap it sounds far more fun!

Well, mimi means 'ear' in Japanese, so... :cat:

PR #1472, opened 2023-11-19
Add a callback for sysex messages in scripts

etene:
This allows python scripts to define a `on_sysex(data)` method that get called by the MidiController much like `on_midi`.

It's called for every sysex MIDI event and takes the raw payload as bytes.

I still have to fix some things before this can be merged:

- [ ] Add documentation
- [x] Prevent arbitrary code injection
- [ ] Fix linting issues

### DIFF
--- a/Source/MidiController.cpp
+++ b/Source/MidiController.cpp
@@ -473,8 +473,13 @@ void MidiController::OnMidiPitchBend(MidiPitchBend& pitchBend)
 
 void MidiController::OnMidi(const MidiMessage& message)
 {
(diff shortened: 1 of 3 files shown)


afentoe:
Hi @etene, thanks for creating this sysex handler.

I have improved a few things and had to make a new pullrequest https://github.com/BespokeSynth/BespokeSynth/pull/1726

I use the on_sysex to parse a program dump from a synthesizer.

NoxyNixie:
I believe this is no longer necessary because https://github.com/BespokeSynth/BespokeSynth/pull/1726 has been merged.

Thanks either way ;)

etene:
I'm terribly sorry, I just saw the notification and I had completely forgotten about this PR...  Thanks for the feedback anyway, happy to know that's been implemented !

PR #1488, opened 2023-12-20
Disconnect controls when midicontroller is disabled

sbungartz:
When a `midicontroller` is disabled (with the circle in the top left corner) it does not accept any inputs but state is still reflected and grids still set lights on pads.

My expectation would be that when a `midicontroller` module is disabled it acts as if it was not there at all, so the lights should not be affected by connections on the controller.
My use case for this: This would allow easily having multiple `midicontroller` instances in various prefabs to control different modules or groups of them, and then using a global `selector` to set the currently controlled module.
But I feel like this would just be more along the lines of what one expects when the `midicontroller` is disabled.

I have been working around this so far by leaving Page 0 empty and switching to that when I wanted to disable a `midicontroller`, so this is kind of where this PR is coming from: Act like we have selected a page with no connections by
- Calling `SetEntirePageToZero(mControllerPage)` to clear the visible state
- Restricting the places where new connections are setup after a page change to only occur when `mEnabled`

### DIFF
--- a/Source/MidiController.cpp
+++ b/Source/MidiController.cpp
@@ -1438,7 +1438,7 @@ void MidiController::GetModuleDimensions(float& width, float& height)
 
 void MidiController::ResyncControllerState()
 {
-   if (mControllerPage >= 0 && mControllerPage < mListeners.size())
+   if (mEnabled && mControllerPage >= 0 && mControllerPage < mListeners.size())
    {
       for (auto i = mListeners[mControllerPage].begin(); i != mListeners[mControllerPage].end(); ++i)
          (*i)->ControllerPageSelected();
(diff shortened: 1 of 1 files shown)


Source/MidiController.cpp
```diff
@@ -1995,7 +2007,7 @@ void MidiController::PostRepatch(PatchCableSource* cableSource, bool fromUserCli
 {
    for (auto* grid : mGrids)
    {
-      if (cableSource == grid->mGridCable)
+      if (mEnabled && cableSource == grid->mGridCable)
```
sbungartz, line 2012: This was necessary since otherwise the `grid->mGridControlTarget[mControllerPage]` was set to `nullptr` from `cableSource-GetTarget()`. Then the grid connection would be lost.

Source/MidiController.cpp
```diff
@@ -1438,7 +1438,7 @@ void MidiController::GetModuleDimensions(float& width, float& height)
 
 void MidiController::ResyncControllerState()
 {
-   if (mControllerPage >= 0 && mControllerPage < mListeners.size())
+   if (mEnabled && mControllerPage >= 0 && mControllerPage < mListeners.size())
```
sbungartz, line 1441: Here I'm not exactly sure if that should still occur or not (since I'm not sure what the `mListeneres` are.

NoxyNixie: The script module can be a listener for instance and I do think they should be triggered when you disable the `midicontroller` so this looks correct.

PR #1489, opened 2023-12-20
Add `pitchtovalue` module

merkys:
This PR introduces a simple `pitchtovalue` module which passes raw MIDI pitch (MIDI note number) as modulation value.

I have a 4x4 button pad which emits MIDI notes. My initial idea was to use it in combination with `pitchtovalue` to control chord types in dropdown list of `chorder` module, but somewhy this does not work as expected (perhaps dropdown receives from it -9999 or 9999 value only?). Anyway it might be useful in other contexts.


merkys:
I do not quite get why CI checks are failing. There seem to be three failures, but when I open them to view I can only find code formatting check failing. I do not quite get why as I tried to mimic code formatting of other modules.

merkys:
Thanks a lot for merging this in, @awwbees!

PR #1527, opened 2024-01-18
Add `chordbounds` module

merkys:
This PR introduces <s>`noteminmax`</s> `chordbounds` which selects minimum- and maximum-pitched notes out of all the notes playing at the same time. This might be used to generate a bass line for notes/chords played with MIDI keyboard.


Source/NoteMinMax.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+   mDestinationCables[0] = new AdditionalNoteCable();
+   mDestinationCables[0]->SetPatchCableSource(new PatchCableSource(this, kConnectionType_Note));
+   AddPatchCableSource(mDestinationCables[0]->GetPatchCableSource());
+
+   mDestinationCables[1] = new AdditionalNoteCable();
+   mDestinationCables[1]->SetPatchCableSource(new PatchCableSource(this, kConnectionType_Note));
+   AddPatchCableSource(mDestinationCables[1]->GetPatchCableSource());
+
+   GetPatchCableSource()->SetEnabled(false);
```
NoxyNixie: I think it might be better to use the "main" patchcable because things like scripts when using the set_target set this one. I'm also not seeing any iteration happening so making `mDestinationCables` an array seems fruitless. Perhaps add a `mSecondaryCable` instead.

Source/NoteMinMax.cpp
```diff
@@ -0,0 +1,150 @@
...
+void NoteMinMax::DrawModule()
+{
+   if (Minimized() || IsVisible() == false)
+      return;
+
+   DrawTextNormal("min", 3, 15);
+   DrawTextNormal("max", 3, 30);
+
+   mDestinationCables[0]->GetPatchCableSource()->SetManualPosition(35, 10);
+   mDestinationCables[1]->GetPatchCableSource()->SetManualPosition(35, 25);
```
NoxyNixie: I believe this can be done on the "main" patchcable too so should not cause issues with my previous suggestions.

Source/NoteMinMax.cpp
merkys: I agree. But could you perhaps point me an example where something similar is done? I looked into `BoundsToPulse` which is similar, but it deals with pulses, not notes.

NoxyNixie: The [send](https://github.com/BespokeSynth/BespokeSynth/blob/main/Source/AudioSend.cpp#L52) module has two outputs one being the "main"/original and a new one added.

NoxyNixie:
I'm going to be annoying ;)

notegate (@PowerUser64) mentioned the name `chordbounds` and I think that name makes more sense as it implies that it acts on multiple notes and states what the outputs are. Could you rename this?

Source/NoteMinMax.cpp
merkys: Fixed in 432cdbaaab6ee1d14a8ec6dca14635338a2f26e2, marking as resolved.

Source/NoteMinMax.cpp
merkys: Right, I have changed the code according to your comment in 432cdbaaab6ee1d14a8ec6dca14635338a2f26e2. Marking as resolved.

merkys:
@NoxyNixie I have just renamed the module to `chordbounds`.

NoxyNixie:
Awesome, I'll see if I can review and merge it soonish ;)

merkys:
What is strange (and I feel might be blocking the merge), is that even after all note off messages are sent, both left and right output cables are shown as lit. Inspecting with `notedisplayer` shows notes with velocity of 0, could they be unpaired note off messages?

NoxyNixie:
ugh eehhmm not sure what happened there. First time trying to push to a pull request.

merkys:
I have solved the issue with missing note off messages. However, there are still some unpleasant sound artifacts appearing at the beginning and the end of a chord. I think they are caused by a signal race between incoming note messages.

merkys:
I have resolved the issue with artifacts appearing at the beginning and the end of chord. They were indeed due to `PlayNote()` being invoked for each note of a chord in the order of increasing pitch. I fixed that by emitting the notes in `OnTransportAdvanced()`. This, however, introduces some delay.

merkys:
Solved the issue with introduced delay - yay!

**Update**: Something else broke in 3a87366: releasing individual notes of a sounding chord will not affect the min and max outgoing notes from the module - will give it a look.

Limitation: notes created by "trigger" button of `notecreator` will not affect the module, see #1916 for issue with the same underlying cause.

PowerUser64:
Nice work! Just tested this out. It seems like there's an issue where the module doesn't output more than one note when note on events occur one at a time. Here's a screen recording of what I mean:

[bespoke-chordbounds.webm](https://github.com/user-attachments/assets/bd35a62d-4b22-4f90-8b88-bc48008ed76a)

Here's the save that I'm using in that recording, if you want to see it for yourself: [bespoke-chordbounds-test.bsk.zip](https://github.com/user-attachments/files/23045626/bespoke-chordbounds-test.bsk.zip)

Try disabling the chordholder module, and you'll see that the module starts working correctly after the first note off event. In order to work around this, two or more notes need to be played simultaneously at the beginning of the sequence of notes.

This is going to be a fun module, I'm looking forward to when it's complete.

merkys:
Thanks for checking this out! Your savestate highlights at least a couple of problems:
* lower bound is not properly output
* note-on and note-off messages are emitted with some delay, which is not nice

I will tend to fixing them.

merkys:
> * lower bound is not properly output

This seems to be caused by both cables from `chordbounds` being fed into the same target. Thus note-off messages coming from one cable terminates note-on messages from the other, which is completely counterintuitive :sweat_smile:

I do not think I can make the behavior dependent on the number of targets. Note streams might converge anywhere downstream, and at that convergence point note-on and note-off messages will interfere. Maybe using separate voice ids would do the trick, but this solution will still produce counterintuitive artifacts for users expecting all notes having voice id of -1.

PR #1528, opened 2024-01-19
Rewrite `pitchtovalue` through `valuesetter` in order to set verbatim values for drop-down lists

merkys:
This PR fixes an issue with `pitchtovalue` and drop-down lists. Prior to this PR, it was impossible to properly control the chord in `chorder` module (for example). In this PR I have rewritten `pitchtovalue` closer to `valuesetter`, thus now verbatim values are set to targets pointed by `pitchtovalue` patches.


merkys:
Thanks a lot @NoxyNixie!

PR #1529, opened 2024-01-19
trigger OBS build on push only

KottV:
This should prevent rebuilds and unnecessary OBS checks on any GH event except Push.


awwbees:
thank you!

PR #1544, opened 2024-03-07
Fix setting patch target (via scripts) not working for some modules

### LINKED ISSUE #1539: Cannot set envelope's target from a script
Hello! I have been trying to create a script for automatically spawning voice chains for some sophisticated polyphony mechanism. Along the way, I have run into some unfortunate limitations of the scripting system. I've also run into a bug that ultimately prevented even a stripped down version:
When I try to set the target of envelope module, either via UI `target` property, or `set_target` script function, it doesn't work, although the `get_target(s)` methods return correct values.
I confirmed that it works, for example, with `curve`.

## Build info
I'm running the latest release version.
`bespoke 1.2.1 (Sep 16 2023 20:17:16) - Linux - HEAD 9a2027e`

## Steps To Reproduce
- Create an `envelope` module & a `script` module
- Add `module.get('envelope').set_target('gain~gain')` into the script
- Run the script

A cable from the envelope to gain slider should appear; envelope should be functionally attached to the destination.
Neither of the cases occur.

## Some discussion
I initially wanted to create a script orchestrating the creation & removal of chains of modules for playing individual voices upon note input. I initially wanted to use prefabs for preparing the skeleton, but prefabs aren't scriptable. So I resorted to manually creating the modules. When I tried to connect the output of a controller (envelope) to some parameter, I ran into a problem that I can't specify the target via a module object, so I had to use a string path. Annoyingly there's no function to get the name of a module, so I had to either: choose the name in the script, or go through the list of all modules returned by `get_modules` and find the target module. I chose the former. I picked a unique generated name with a set of used names, and concatenated the parameter name to get a path.
It would help if there were more APIs for this kind of workflow.

dekrain:
This should fix #1539. I had it fixed for 3 days, but I'm just now making the PR.
It might be worthwhile extending the functionality by allowing to select which source to patch (or add a new one), or to change the main source (if that doesn't break anything), but for now it suffices.

### DIFF
--- a/Source/IDrawableModule.cpp
+++ b/Source/IDrawableModule.cpp
@@ -551,7 +551,8 @@ void IDrawableModule::SetTarget(IClickable* target)
 {
    PatchCableSource* source = mMainPatchCableSource;
    size_t idx = 0;
-   while (source == nullptr || !source->Enabled()) {
+   while (source == nullptr || !source->Enabled())
+   {
       // Find first patch cable in the list that's enabled.
       if (!(idx < mPatchCableSources.size()))
          return;
--- a/Source/IDrawableModule.cpp
+++ b/Source/IDrawableModule.cpp
@@ -1366,3 +1366,13 @@ bool IDrawableModule::IsSpawningOnTheFly(const ofxJSONElement& moduleInfo)
       return false;
    return moduleInfo["onthefly"].asBool();
 }
+
+PatchCableSource* IDrawableModule::GetPatchCableSource(int index)
+{
+   if (index == 0)
+   {
+      if (mMainPatchCableSource != nullptr || mPatchCableSources.empty())
+         return mMainPatchCableSource;
+   }
+   return mPatchCableSources[index];
+}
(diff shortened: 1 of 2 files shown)


NoxyNixie:
I've been working on an alternate way of fixing this which is a bit more involved but I can likely add your enabled check to my patch too.

dekrain:
Okay! Can you share more details about what you're working on/how you're approaching it?

dekrain:
I also moved the `GetPatchCableSource` function to the `cpp` file, so that it can be modified without recompiling most of the project. (I tried fixing my issue by changing it, but ended up not having to.)

NoxyNixie:
I'm removing `mMainPatchCableSource` completely since it no longer makes sense. I'm making the modules that don't use their primary patchcablesource remove the original one instead of hiding them (there is no point for this data to stick around) and because of this I need to take care of loading old savestates correctly.

While testing I've been running into all kinds of other issues (as usual) that are completely unrelated to my changes.

I would work on it now but I'm having a migraine so maybe tomorrow or this weekend I get to actually making it a PR

NoxyNixie:
Alrighty my approach posted as PR, feel free to leave feedback ;)

awwbees:
root issue resolved in #1548

PR #1545, opened 2024-03-07
Fix the default for modwheel.

NoxyNixie:
The default for modwheel should be 0 since this is the "neutral" position of the modwheel. When you push the modwheel to hte max this value will become 2 and when pulling to the min it will become -2.

`0.5` is therefore a incorrect default.

### DIFF
--- a/Source/ModulationChain.h
+++ b/Source/ModulationChain.h
@@ -73,7 +73,7 @@ struct ModulationParameters
    float pan{ 0 };
 
    static constexpr float kDefaultPitchBend{ 0 };
-   static constexpr float kDefaultModWheel{ .5f };
+   static constexpr float kDefaultModWheel{ 0 };
    static constexpr float kDefaultPressure{ .5f };
 };
 


awwbees:
there's a complicated history with this one that I've gone back and forth on, I need to find the time to dig in and remember the details. iirc the range of modwheel should just be 0 to 1.

NoxyNixie:
> iirc the range of modwheel should just be 0 to 1.

Well it isn't and the modules that deal with modwheel have incorrect behavior right now. Like if you just spawned the `modwheel` module you must wiggle its slider before it can set it to `0` even though it spawns in with `0`. Yes this can likely be fixed in the `modwheel` module but this is not the only module that suffers from this. All plugins also suffer here as they will by default now receive 25% modulation (`0.5`). As far as I've seen in the code, most if not all code assumes `0` to be no modulation.

ArkyonVeil:
After much trial and error with #1550 , I've reached the same conclusion as this PR. This should fix Omnisphere and perhaps other VSTs which expect the same baseline.

Edit: Made the same modification locally and can confirm that it's fixed.

caesarbear:
Having any modulation by default doesn't make any sense. I'm always assuming it's at zero when I start. Some projects can have a lot of instruments. Real pain to have to jiggle the mod wheel on all of those when I load them in.

afentoe:
The `notelooper` has this annoying behavior as well: after recording a loop, it will add CC Modulation messages to each note, with a value of 3F (0.5), resulting in unexpected effects on connected synths and VSTs.

So +1 from me, to accept this pull request

PR #1548, opened 2024-03-14
PatchCableSource tweaks for a couple modules. Resolves: #1539.

### LINKED ISSUE #1539: Cannot set envelope's target from a script

NoxyNixie:
PatchCableSource tweaks for a couple modules. Resolves: #1539.

- Made it possible to set the target using scripts for the modules: `audiotocv`, `envelope` and `leveltocv`.
- Added a new `set_target` script method with an additional `cableSourceIndex` parameter. 
- Removed hidden main patchcablesource from the modules: `audiotocv`, `envelope` and `leveltocv`.
- Removed the use of a `mMainPatchCableSource` cleaning up the code a bit.

This uses some of the suggestions made in #1544 but is not compatible with that PR (they conflict).
Resolves the same bug but with a multifaceted approach and some code cleanup of runtime data.

### DIFF
--- a/Source/AudioLevelToCV.cpp
+++ b/Source/AudioLevelToCV.cpp
@@ -54,8 +54,6 @@ void AudioLevelToCV::CreateUIControls()
    FloatSliderUpdated(mAttackSlider, 0, gTime);
    FloatSliderUpdated(mReleaseSlider, 0, gTime);
 
-   RemovePatchCableSource(GetPatchCableSource());
-
    mTargetCable = new PatchCableSource(this, kConnectionType_Modulator);
    mTargetCable->SetModulatorOwner(this);
    AddPatchCableSource(mTargetCable);
--- a/Source/AudioLevelToCV.cpp
+++ b/Source/AudioLevelToCV.cpp
@@ -54,7 +54,7 @@ void AudioLevelToCV::CreateUIControls()
    FloatSliderUpdated(mAttackSlider, 0, gTime);
    FloatSliderUpdated(mReleaseSlider, 0, gTime);
 
-   GetPatchCableSource()->SetEnabled(false);
+   RemovePatchCableSource(GetPatchCableSource());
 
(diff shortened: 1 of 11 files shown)


dekrain:
Adding back usage of `mMainPatchCableSource` in the second commit wasn't actually necessary. Since nothing is setting it anymore, all code paths checking it are redundant, and it makes `SetUpPatchCables` always crash, since it requires it to be set. When instantiating some modules, it crashes immediately, including when loading stored patches.

PR #1552, opened 2024-03-18
Remove mMainPatchCableSource entirely

dekrain:
As mentioned in https://github.com/BespokeSynth/BespokeSynth/pull/1548#issuecomment-2004061003, reintroduction of `mMainPatchCableSource` wasn't necessary, as it wasn't set anymore. This also fixes a crash when calling `SetUpPatchCables` because it required `mMainPatchCableSource` to be set.


NoxyNixie:
Thanks was about to make this exact PR ;)

Issue #1551, opened 2024-03-18
Feature - Assign MIDI channel to notes


geoffeep:
## Summary

When passing notes into a VST it would be great to support setting a MIDI Channel (multi instrument VSTs like Kontakt can use MIDI channels to route specific notes to specific instruments).

For example, if I have a Kontakt VST set up with multiple Random Note nodes I'd like those to each play a different Kontakt instrument with just a single Kontakt node placed (this is handled through MIDI Channels in Kontakt and Reaper).

![image](https://github.com/BespokeSynth/BespokeSynth/assets/7421471/69534438-bc4d-4c72-a9aa-d2889a28f5e4)

### Motivation

When I use Kontakt in Reaper I have a single instance of the VST that can load many instruments into a multi-rack (for example "Textural Violin" and "Gentle Winds" free libraries, one set to MIDI Channel 1 for input, the other set to MIDI Channel 2):

![image](https://github.com/BespokeSynth/BespokeSynth/assets/7421471/79eca8a0-6b34-4d6f-8320-7b1a2a19d10c)

Even when working with 4 different instruments at once, fed by 4 different MIDI channels in Reaper, CPU usage doesn't get much higher than 10% using this approach.

In Bespoke to replicate the same setup I have to create a separate Kontakt VST node for Kontakt 4 times, and my CPU usage goes up about 20% for each instance of that node, so I really can't use more than 2-3 Kontakt instruments before I start getting audio issues:

![image](https://github.com/BespokeSynth/BespokeSynth/assets/7421471/f498cf85-07ac-4c78-8dfa-d20c06f401ec)

ArkyonVeil:
Just a heads-up to help you out until there's a cleaner way to do this. It feels like a bit of a hack, but it works.

Funnel your red instruments through the orange effect "Portamento", and then configure its Voice_IDX (this will be your channel)
Then in the VST tick the usevoiceaschannel checkbox. Should be good to go.
![image](https://github.com/BespokeSynth/BespokeSynth/assets/42283444/86ba81bc-0ed2-42a1-87b9-c4b4b61f3610)

dekrain:
There's a parameter called _voice index_ that's passed around modules that produce/consume notes, although it isn't used much (only use I found are in `midicontroller` as an option to map **channel IDs** to voice idx) and most often it's set to _-1_. It should be viable to create a module that sets a desired voice ID for incoming notes, and then to tell the VST module to map voice IDs back to MIDI channels.

[...]

PR #1553, opened 2024-03-20
Set Note Voice node and debug build VST assert fix

geoffeep:
[https://github.com/BespokeSynth/BespokeSynth/issues/1551]

I have created a new node called Note Set Voice that allows manual setting of a voice. Combined with UseVoiceAsChannel on VST plugins this is really helpful, because some plugins like Kontakt allow you to specific a midi channel input for each individual instrument, so this allows me to use just one Kontakt node to play many instruments (which is helpful because Kontakt is quite CPU heavy).

There was also an issue in VSTPlayhead that was causing an assert in debug builds, so I wrote some checking around the optional timesignature variable


geoffeep:
One thing I should probably work out before submitting this: I changed the documentation json, but am unsure how to regenerate the html, which I think I need to do to see my tooltip? Unless that automatically gets built somehow?

NoxyNixie:
> One thing I should probably work out before submitting this: I changed the documentation json, but am unsure how to regenerate the html, which I think I need to do to see my tooltip? Unless that automatically gets built somehow?

Tooltips are in https://github.com/BespokeSynth/BespokeSynth/blob/main/resource/tooltips_eng.txt#L1595

PR #1565, opened 2024-03-30
Make it so floatsliders compute their values using the audiopoller instead of the render method. Resolves: #1061.

### LINKED ISSUE #1061: Bug - metronome does not respond to modulation when off screen
## Describe the bug

Some modules doesn't respond to automation unless they are visible.

## Build info

3fc87549496bcda1650a37f166ba3b4e08260067

## Steps To Reproduce

1. Create a metronome and connect it to the main output
2. Attach a slow LFO to the metronome
3. Listen to the metronome's sound change in volume as it plays
4. Pan the view away from the metronome such that it isn't in sight anymore
5. Notice how the metronome's volume level does not change

🔊🔊🔊 sound on ⬇️⬇️⬇️

https://github.com/BespokeSynth/BespokeSynth/assets/24578572/f8be24de-9241-4ee0-9076-b29d07b8102e

This was also documented happening with the midicc module [by @NotSoAnon on discord](https://discord.com/channels/846834873254805554/854199359913328671/1223431078140968985).

[Discord_HQ.webm](https://github.com/BespokeSynth/BespokeSynth/assets/24578572/0af06c84-bab6-434c-95e5-811d406affb2) 

The metronome should respond to automation whether it is in view or not. Right now, this is not the case.

NoxyNixie:
Make it so floatsliders compute their values using the audiopoller instead of the render method. Resolves: #1061.

This fixes offscreen floatslider not being properly computed. Affects almost all modules with floatsliders.

### DIFF
--- a/Source/Slider.cpp
+++ b/Source/Slider.cpp
@@ -61,14 +61,15 @@ FloatSlider::FloatSlider(IFloatSliderListener* owner, const char* label, IUICont
 
 FloatSlider::~FloatSlider()
 {
-   if (mIsSmoothing)
-      TheTransport->RemoveAudioPoller(this);
+   TheTransport->RemoveAudioPoller(this);
 }
 
 void FloatSlider::Init()
 {
    if (mVar)
       mOriginalValue = *mVar;
+
+   TheTransport->AddAudioPoller(this);
 }
 
 FloatSliderLFOControl* FloatSlider::AcquireLFO()
@@ -104,9 +105,6 @@ void FloatSlider::SetModulator(IModulator* modulator)
 
 void FloatSlider::Render()
 {
-   if (mLastComputeTime + .1f < gTime)
-      Compute();
-
    float normalWidth = mWidth;
    float normalHeight = mHeight;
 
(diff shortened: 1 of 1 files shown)


awwbees:
are you able to do any before-and-after performance testing on this one? the overhead of having every floatslider always be audio-polled has me a bit wary, it's a lot more work on the audio thread. I do agree that this is solving an important problem, but I wonder if there's another approach, like only adding sliders as audiopollers if a modulator becomes attached. (although with this approach there is a lot more bookkeeping and potential for bugs, unfortunately...)

NoxyNixie:
I've not done extensive performance testing but I did test with with [LAM Artificial savestate](https://discord.com/channels/846834873254805554/853305813015593011/1212519480291368960) which to me seemed to run at similar performance before and after the patch.

When I didn't notice a significant performance hit on that one I decided to no not look deeper as the benefit to mee seems to outweighs the performance hit (if there is one).

awwbees:
I'll find some time to measure performance before pulling this in

NoxyNixie:
I did a crude test with LAM-artificial and found that it is very close in performance. Unfortunately it is not a perfect comparison as it relies on my human input for how quickly I load/stop BespokeSynth but it shows me enough information to know that the difference (if there actually is one) is very minimal.

Before patch:
devenv_20240416_144649

After Patch:
devenv_20240416_143557

These are both done in release mode. Note that the "before patch" one is slightly shorter which if you calculate that in means the resulting values are nearly identical.

Both tests ran at about 22fps and I did not move the viewport around (mostly had my mouse still as well so to not add more randomness).

awwbees:
there's an important distinction here, which is that we're changing where Compute() happens from the poll thread (which is mostly OK if it's slow) to the audio thread (which results in audio dropout/stutter/bad sounds if it's slow), so the profiling needs to be a little more specific than looking at general CPU usage.

awwbees:
I ended up fixing this via #1631 instead, so the overhead of computing on the audio thread is still determined by the slider's owning module. this approach will have identical performance to the previous approach in the case where all sliders are onscreen

Issue #1466, opened 2023-11-14
Target Framerate


ArkyonVeil:
## Summary

Simply put, an option in the settings to request 60 fps, 30 fps (for say power saving), or higher such as 144 for those with the hardware that can handle it.

### Motivation

It would help users modulate the level of performance they desire or maintain a similar level of visual smoothness throughout their OS and DAW experience.

PR #1567, opened 2024-04-02
Adds the ability for the user to configure their visual Frame Rate

ArkyonVeil:
Handy for a smoother appearance on high refresh rate monitors, or also for squeezing more CPU cycles out of an intense music session.

Default frame rate remains 60.

Wishlisted on #1466

https://github.com/BespokeSynth/BespokeSynth/assets/42283444/ad41705a-1967-44b4-9f0c-aaca1c6df477

### DIFF
--- a/Source/UserPrefs.h
+++ b/Source/UserPrefs.h
@@ -287,9 +287,9 @@ public:
 
    static std::string ToStringLeadingZeroes(int number);
 
+   float LastTargetFramerate;
    ofxJSONElement mUserPrefsFile;
    std::vector<UserPref*> mUserPrefs;
-
    UserPrefDropdownString devicetype{ "devicetype", "auto", 200, UserPrefCategory::General };
    UserPrefDropdownString audio_output_device{ "audio_output_device", "auto", 350, UserPrefCategory::General };
    UserPrefDropdownString audio_input_device{ "audio_input_device", "none", 350, UserPrefCategory::General };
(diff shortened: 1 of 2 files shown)


Source/UserPrefs.h
```diff
@@ -330,11 +330,11 @@ class UserPrefsHolder
    UserPrefFloat background_r{ "background_r", 0.09f, 0, 1, UserPrefCategory::Graphics };
    UserPrefFloat background_g{ "background_g", 0.09f, 0, 1, UserPrefCategory::Graphics };
    UserPrefFloat background_b{ "background_b", 0.09f, 0, 1, UserPrefCategory::Graphics };
+   UserPrefFloat target_framerate{ "target_framerate", 60, 30, 144, UserPrefCategory::Graphics };
```
NoxyNixie, line 333: hhmmm, I'd make this lower limit 1FPS and perhaps the upper limit much higher as well.

ArkyonVeil: I don't really see much point in allowing the framerate to go as low as 1FPS. This is only the target framerate, so if the session weighs too much on the rendering, it will go lower regardless. Maybe 240 fps could be the new limit as there's monitors out there with that refresh rate.

Edit: A better solution for high frame rates would probably be V_Sync, but that sounds way harder than an adjustable number.

NoxyNixie: Well my reasoning for 1FPS is that it is more than 0 and probably works with JUCE. There are situations where framerate is not at all important (i.e. when you are performing the patch without even looking at the screen, or live) but you'd rather have it all available for audio processing.

Or when you are on a low performance system and would rather have more CPU cycles dedicated to audio processing as opposed to rendering (yes rendering uses quite a bit of CPU).

Beyond that why not give the power to the user? Maybe someone has a reason to go low FPS that we haven't thought of, in such a case I'd rather be an enabler than a limiter.

ArkyonVeil: Using Bespoke at 1 FPS or below seems like the territory of a headless environment, or when offscreen, IE where the UI doesn't have to render at all. And the user can still use Bespoke somewhere else freely. Otherwise rendering at really low framerates will just make it nigh unusable.

My decision was simply because 30 FPS is a nice ballpark measure of remaining usable enough while being somewhat choppy. Mainly for ensuring the UI starts and ends at a satisfying number. And the UI is responsive enough to not feel broken.

Ideally I'd like to cap the maximum value at the refresh rate of the user's monitor. But as mentioned, it's harder to do properly.

> Beyond that why not give the power to the user? Maybe someone has a reason to go low FPS that we haven't thought of, in such a case I'd rather be an enabler than a limiter.

Not sure if such a user that really wants to limit their framerate to the single digits even exists. Or at least not a user that is using Bespoke normally.

And if they really care all that much? They can figure out a way...
image
In Bespoke, everything is a module, even the settings.

dekrain: Now I wonder: what happens at negative framerate?

ArkyonVeil: > Now I wonder: what happens at negative framerate?

Freezes.

PR #1568, opened 2024-04-02
Add memory block & custom stream API to `FileStream`s

dekrain:
It makes `FileStreamInput/Output` APIs use base `juce::Input/OutputStream` classes instead of file-specific ones, so that you can serialize from/to `MemoryBlock`s (and other streams).
As a result, using a temporary file when duplicating modules (!) and using Snapshot is no longer necessary.
It also allows my temporary implementation for a `PolyphonicContainer` to duplicate modules fast without interacting with the OS.

### DIFF
--- a/Source/FileStream.cpp
+++ b/Source/FileStream.cpp
@@ -29,7 +29,8 @@
 //static
 bool FileStreamIn::s32BitMode = false;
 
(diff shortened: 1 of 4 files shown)


awwbees:
nice!

PR #1569, opened 2024-04-03
Added mouse click based velocity support to KeyboardDisplay

ArkyonVeil:
In action:

https://github.com/BespokeSynth/BespokeSynth/assets/42283444/611c58bc-0655-423a-95e3-372d7c90029c

Should work with keyboards of any size, velocity depends on the y value of the click position. Where most of the white area being where the clicks are registered at full power.

PS: Ignore the white outline, it's something else I'm working on and not present in this PR.

### DIFF
--- a/Source/KeyboardDisplay.cpp
+++ b/Source/KeyboardDisplay.cpp
@@ -103,7 +103,7 @@ void KeyboardDisplay::OnClicked(float x, float y, bool right)
                   else
                      kOffsetVal = mHeight;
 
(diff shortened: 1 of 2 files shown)


ArkyonVeil:
Seriously? I reformatted the code twice.

Hmm, have to check the custom clang format then. :[
image

Edit: Done and Rider can't seem to load it.

Edit2: Tried a dozen approaches, half in Rider. Another half by downloading LLVM 17.0.1 and running a batch file over the ENTIRE source code. There were no changes over ANY file. Even intentional mistakes are "corrected" to what the formatter declared an "error" I am officially out of ideas.

PR #1570, opened 2024-04-05
KeyboardDisplay Overhaul

ArkyonVeil:
KeyboardDisplay module overhaul.
- Automatic octave rescaling, your keyboard will try to fit more octaves as it has room on it.
- Hover based keyboard capture, try typing with your mouse hovering over a keyboard!
- Octave labels
- Jump octaves via keyboard keybinds
- Added a Keyboard2Midi layout, based on the current one, which now named "Ableton"
- Added a second layout "Fruity" inspired by the FL Studio layout with stacked octaves
- Added a third layout "ignore" which just stops hover behavior capture entirely.
- Responds better to being enabled/disabled
- Probably quite a few fixes when messing with what used to be module settings options
- Added an option in the settings to hide labels. For when you want that clean look.
- Also another option that enforces a fixed number of octaves regardless of the keyboard size. Which prevents the automatic rescaling altogether.

https://github.com/BespokeSynth/BespokeSynth/assets/42283444/1e59ae68-29d3-413a-8697-bdd2be450990


awwbees:
thanks, this is great. I've got a few minor issues before accepting, I'll make some comments inline

Source/HoverSelectModule.cpp
```diff
@@ -0,0 +1,57 @@
...
+bool HoverSelectModule::MouseMoved(float x, float y)
+{
+   float mWidth;
+   float mHeight;
+   GetModuleDimensions(mWidth, mHeight);
+   if (x > 0 && y > -+mTitleBarHeight && x < mWidth && y < mHeight + mTitleBarHeight)
+   {
+      if (!hovered && !TheQuickSpawnMenu->IsShowing())
+      {
+         if (!IsEnabled() || UserPrefs.keyboard_2_midi_layout.GetIndex() == 2)
```
awwbees: I like how HoverSelectModule is a nice generic class, but I think this generic nature of it gets broken by it caring about a keyboard-related user preference. is there a different way to handle this?

Source/Keyboard2MidiLayout.h
```diff
@@ -0,0 +1,21 @@
+﻿#pragma once
+
+enum class Keyboard2MidiLayoutType
+{
+   Ableton,
+   Fruity,
+   Ignore
```
awwbees: maybe this should be referred to as "ignore hover", or something... "ignore" is a strange name without the context

Source/QuickSpawnMenu.h
```diff
@@ -51,7 +51,7 @@ class QuickSpawnMenu : public IDrawableModule
 public:
    QuickSpawnMenu();
    virtual ~QuickSpawnMenu();
-
+   inline static bool enableKeyboardInput = true;
```
awwbees: the statefulness here seems bug-prone, I think we should remove this variable. maybe instead, it should be checking to see if there's any active HoverSelectModule that wants to be consuming keyboard input.

Source/HoverSelectModule.cpp
ArkyonVeil: Not really sure how to work around this. It's meant to listen to PC keyboard inputs and read them from their preferred layout that they bring in from other DAWs.

While we can make this layout config per module, it's kind of like when load up PhotoshopBespoke, you have one part of the UI with Photoshop layout, while the default every brush spawns with is the GIMP layout.

The user just to jump mental hoops to see which module expects which layout, and have to set their preferred setting every time.

The right way to fix this is allow the user to allow the user to customize their keybinds, and then over the Hover Module, they can pick which layout they like. This way they can probably control multiple modules from one.

But right now it doesn't make much sense since both Fruity and Ableton layouts are rather similar.

---

Some other thoughts on the Generic hover class, it's correct name would probably be more like "HoverKeyboardCaptureModule" since it's tightly related to when a user is "hovering" over it, to capture their keyboard. Don't really see much of a reason for other modules to hold this behavior, outside maybe the VST one. And those will probably expect keyboard keybinds anyway as well as alternative inputs from the other MIDI centered modules.

Later on I'd like to decouple most of the keyboard logic from the KeyboardDisplay over the hover module, so this generic class is more drag and drop so to speak. It works but its a bit fiddly right now.

Another thing that I'd like to throw in is a toggle like this, so that you could continue type on the keyboard and the module would listen while the mouse is anywhere.
Photo_oDd85BLA1m

Source/Keyboard2MidiLayout.h
ArkyonVeil: Can also write "none", which given the context of settings description would make reasonable sense.

Source/QuickSpawnMenu.h
ArkyonVeil: Right, seems like a good idea. I'll take a look.

Source/HoverSelectModule.cpp
awwbees: I think I've got a good idea to handle this, it would involve allowing the current hover module to intercept key input and consume it or pass it along, based upon the key. add a method `virtual bool ShouldConsumeKey(int key) { return false; }` to HoverSelectModule, then implement it in KeyboardDisplay. that way KeyboardDisplay could look at that user pref and return `true` if the user is pressing a key within the keyboard2midi mapping, and HoverSelectModule wouldn't have to care about the user pref. I'm happy to take a look at this, unless you'd prefer to handle it yourself.

PR #1573, opened 2024-04-12
Added major thirds layout

ftredget:
Added the major thirds layout to the LaunchpadKeyboard as it is another popular layout on isomorphic / grid keyboards.


ftredget:
would it be possible to add a user-customisable layout for the Launchpad / gridcontroller where you can set the intervals for one step along and one step up (as in, user inputs n and m so that the point (x,y) on the grid returns note n\*x+m\*y)?

PR #1586, opened 2024-05-08
Refresh the parameter dropdown list on ParamInfo change

baconpaul:
If (1) Parameter Names change and (2) Parameter Count is less than 30, the code to update parameter names didn't update the in-use mShowParameterDropdown. Correct this omission, and since I needed the length in two spots, move the "30" to "kMaxParametersInDropdown"

Turns out, by the way, that Airwindows Consolidated is a plugin which (1) changes parameter names and (2) has parameter count less than 30. So you know, some enlightened self interest in this PR!


awwbees:
thanks!

PR #1592, opened 2024-06-10
Don't use file path hash for plugins lookup.

KottV:
As it turned out that `createIdentifierString()` [createIdentifierString()](https://forum.juce.com/t/createidentifierstring-is-not-independent-of-the-plug-ins-file-location/58890) is not platform or even file path independent.
So this function adds a second step in plugin lookup (without a file path hash) after trying to find by exact `IdentifierString`.

### DIFF
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -3,5 +3,5 @@ workflow:
     - trigger_services:
          project: home:bespokesynth
          package: bespokesynth-nightly
-    - filters:
+  filters:
          event: push
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -3,3 +3,5 @@ workflow:
     - trigger_services:
          project: home:bespokesynth
          package: bespokesynth-nightly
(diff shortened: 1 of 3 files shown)


awwbees:
this seems useful and I will probably merge it, can you describe more specifically what situation is being handled here?

KottV:
> this seems useful and I will probably merge it, can you describe more specifically what situation is being handled here?

ATM Bespoke's VSTPlugin.cpp uses `createIdentifierString()` for identifying plugins. This string is `FORMAT-name-hash-id`, unfortunately `hash` is path specific (opposite to the function description), so we can't find the plugin from the patch by `id` if it sits in the another directory. So, if we didn't find the plugin from the patch by the  first take, this workaround removes the `hash` and does a second lookup iteration.

I think we should rework the plugin handling, using the XML string. But for now this workaround does the job.

awwbees:
thanks for the explanation! agreed that the real fix should be pursued in the future

PR #1596, opened 2024-06-16
add support for template projects (.bskt)

PowerUser64:
This adds support for template projects, which are the same as a regular project but have the `.bskt` file extension. Attempting to save an open project file that has the template extension will prompt you to save it as a file of the same name, but with a date appended and with the `.bsk` extension instead.

Ex: `/path/my_cool_project.bskt` -> `~/Documents/BespokeSynth/savestate/my_cool_project_2024-06-15_21-02.bsk`


PowerUser64:
Ooh good fix!

PR #1597, opened 2024-06-16
add environment variable for setting bespoke data dir

PowerUser64:
This adds adds an environment variable called `BESPOKE_DATA_DIR` that can be set to modify the location of the bespoke data directory (`~/Documents/BespokeSynth` by default).

### DIFF
--- a/Source/OpenFrameworksPort.cpp
+++ b/Source/OpenFrameworksPort.cpp
@@ -17,11 +17,11 @@
 **/
 /*
  ==============================================================================
- 
+
  OpenFrameworksPort.cpp
  Created: 30 May 2016 9:13:01pm
  Author:  Ryan Challinor
- 
+
  ==============================================================================
  */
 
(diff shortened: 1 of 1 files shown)


Source/OpenFrameworksPort.cpp
```diff
@@ -69,7 +69,8 @@ std::string ofToDataPath(const std::string& path)
 
    static const auto sDataDir = []
    {
-      auto dataDir = juce::File::getSpecialLocation(juce::File::userDocumentsDirectory).getChildFile("BespokeSynth").getFullPathName().toStdString();
+      auto defaultDataDir = juce::File::getSpecialLocation(juce::File::userDocumentsDirectory).getChildFile("BespokeSynth").getFullPathName().toStdString();
+      auto dataDir = juce::SystemStats::getEnvironmentVariable("BESPOKE_DATA_DIR", defaultDataDir).toStdString();
 #if BESPOKE_WINDOWS
       std::replace(begin(dataDir), end(dataDir), '\\', '/');
 #endif
```
PowerUser64, line 76: There's a line just below here that reads "`dataDir += '/';`" - should this line be before the part that changes the forward slashes into backslashes for windows? @awwbees

awwbees: nah, the replacement goes the other way: backslashes turn into forward slashes, and bespoke wants forward slashes for paths everywhere

PowerUser64: ahh I see

PR #1599, opened 2024-06-18
add support for command line options

PowerUser64:
This adds support for the `--help`, `--version`, and `--option` command-line options, as well as their corresponding short options.

Most notably, `--option`, which allows overriding values in `userprefs.json`.

Note that some of the prose in the help menu is written with template (`.bskt`) files in mind, as proposed in #1596.

### DIFF
--- a/Source/Main.cpp
+++ b/Source/Main.cpp
@@ -1,16 +1,17 @@
 /*
  ==============================================================================
- 
+
  This file was auto-generated!
- 
+
  It contains the basic startup code for a Juce application.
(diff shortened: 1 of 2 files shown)


PowerUser64:
Noticed I forgot two newlines on lines 56 and 57 in `Main.cpp`, fix'd.

PowerUser64:
Would it make everything more "bespoky" if I make everything in `--help` lowercase? Or should I leave it uppercase like most other things are?

PR #1600, opened 2024-06-18
Various improvements to git repository and code editing experience

PowerUser64:
- [.editorconfig](https://editorconfig.org) - tells editors how files should look (different from clang-format)
- [.gitattributes: text](https://git-scm.com/docs/gitattributes#_text) - settings for git, here it's being used to convert line endings to the defaults for the host
- .gitignore - ignore `.ccls-cache` folder, created by the ccls language server

Please review the options in `.editorconfig`. I think they should be right, but I'd like to make sure. I chose them based on the contents of `.clang-format` and existing files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+[*]
+insert_final_newline = false
+charset = utf-8
+
+# c, cpp, h, hpp files and .i
+[*.{i,{c,h}{,pp}}]
+indent_style = space
+indent_size = 3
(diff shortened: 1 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,7 @@
+[*]
+insert_final_newline = false
+charset = utf-8
+
+[*.{c,h}{,pp}]
```
NoxyNixie: This should include `.i` for the python bindings.

PowerUser64: Good catch, thanks!

PR #1602, opened 2024-06-19
#pragma once everywhere.

NoxyNixie:
Part of code cleanup, I see no reason why not to use this everywhere, this also makes it way more consistent.

### DIFF
--- a/Source/ADSRDisplay.h
+++ b/Source/ADSRDisplay.h
@@ -23,8 +23,7 @@
 //
 //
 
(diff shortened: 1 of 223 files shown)


awwbees:
agreed! the inconsistency of seeing the "include guard" style is really just because bespoke is old enough that "pragma once" was less widely-used way back then, so that was how I used to do headers (which I think is the style openFrameworks used back then)

PR #1603, opened 2024-06-28
Addressing various "code quality" issues detected by ReSharper and general cleanup of code.

NoxyNixie:
Addressing various "code quality" issues detected by ReSharper and general cleanup of code:

- Possibly unused #include directive
- ﻿Assigned value is never used
- ﻿Declaration of a variable hides previous local declaration
- ﻿Members are not initialized in this constructor
- Deprecation warnings
- Unused (and commented out) code
- Default case is not handled in the switch statement
- ﻿Include guard is not found at the beginning of a header file


NoxyNixie:
Well, I'll be trying to fix the compilation issues on mac and linux later ;)

Source/LinnstrumentControl.cpp
```diff
@@ -27,15 +27,15 @@
...
-#include "IAudioSource.h"
 #include "ModularSynth.h"
-#include "FillSaveDropdown.h"
 #include "ModulationChain.h"
 #include "PolyphonyMgr.h"
 #include "MidiController.h"
-#if BESPOKE_LINUX | BESPOKE_MAC
-#include <unistd.h>
-#endif
+#include "ableton/Link.hpp"
```
awwbees: what's Link.hpp doing here? resharper being weird, or a copy-paste error?

Source/Profiler.cpp
```diff
@@ -39,6 +38,7 @@ namespace
 {
    static inline uint64_t rdtscp(uint32_t& aux)
    {
+      //@TODO(Noxy): Why is windows treated differently here? Doesn't std::chrono::high_resolution_clock::now work on windows?
```
awwbees, line 41: maybe. I went through lots of painful different approaches back when I was making this work, which is possibly unnecessary now

Source/LinnstrumentControl.cpp
NoxyNixie: Looks like a copy pasta failure.

NoxyNixie:
Alright well this was an adventure. Initially tried two different ways of fixing the deprecation warnings on strncpy but the resolution to that one is different on the different platforms we support. I didn't want to add a bunch of conditionals and changing the stuff to std::string made me run into other issues. So I reverted the strncpy changes.

I think this is ready for review/merge now.

awwbees:
there are some changes in here that I want to keep (especially the bugfixes!) and a few that I'm less fond of, so I've gotta figure out the best way to go through all this and roll back the stuff I don't want to merge. I'll probably do it bit-by-bit to make it less daunting

awwbees:
the midi device updates have fixed the deprecation warnings, but seem to have broken midi device support in my testing. midi devices in the midicontroller module all show up in the dropdown, but say "not connected". were you able to test these changes with any midi controllers?

awwbees:
it looks like there's some device name/identifier stuff that needs to be untangled to get midi controllers working again

awwbees:
I pulled the midi device deprecation fixes out, and I'll make a new PR with those, since it's a change with some degree of complexity and is pretty different from the rest of this PR

awwbees:
hmm, not sure what the clang-format error is here, everything looks fine to my eyes and my local clang-format check reports no issues

PR #1645, opened 2024-09-19
Made floatsliders use `double` instead of `float` to allow for increased precision when doing modulation math.

NoxyNixie:
Made floatsliders use `double` instead of `float` to allow for increased precision when doing modulation math.

Consequently a lot of internal math is now more precise especially when it comes to calculations with large numbers like for instance samplerates.

Almost everything is touched so this commit should probably be merged in after a release of for instance 1.2.2 or 1.3.

I think I have caught all loading issues caused by this change, at least all of my old savestates load including the one where I added all the modules.

This was several weeks of work including some automated tasks. It was tested but not yet extensively as one should with a patch like this and I expect to have to adjust things so don't hesitate to ask me to do changes.

Also I am aware that this patch will conflict with just about any other PR currently out there and I won't mind having to resolve conflicts.

For reference:
image

"A little further" was an understatement 😛 

I've not touched the audio buffers since those tend to use vector optimized operations which would have a measurable difference if we make those double. Though I do think having the option for 64bit float audio precision might be interesting in the future (and as "selling point") so with that in mind I made most calculation surrounding the audio buffer use double precision instead of float.

I tried to not stray too far away from my initial goal but there are some minor bugfixes here and there.


PowerUser64:
How will this impact the recording feature? Will saved recordings be twice as big since each sample uses twice as many bits?

NoxyNixie:
I am initially not touching the sample data, it will remain 32bit float for the time being.

The main reason is performance impact. Almost all audio buffer operations use the functions `Mult` or `Add` which has vector optimisations enabled. If we switch data type there the performance will at least be twice as slow if not more. 

However for all the other non vector optimized math it should for the most part not have any impact on performance aside from a minor increase in memory usage (going from 32bit to 64bit floating point values).

I am however also slightly preparing for the possibility of 64bit float sample data in case someone really wants this (I don't know if there are recorders doing this high of a bit depth) but it won't be the goal of this PR. It will be its own PR possibly after this or after some other PR's I want to do as well.

afentoe:
@NoxyNixie 
This bsk file causes audio glitches and full mutes, which are only present in the double branch and work fine in older versions.
It's in the old non-double format, so it can be loaded in both versions for comparison.
I suspect high audio input volumes into the `compressor` module.

[doubling-audio distortion.zip](https://github.com/user-attachments/files/19492783/doubling-audio.distortion.zip)

afentoe:
Combi of `dotsequencer` with `snapshots` will cause Bespoke `double` to crash on loading the savestate. See attached examples.
You can open the `snapshots dotseq - save by nightly` in Bespoke `double`, but when you save the file and load it again, it will crash.
Loading `snapshots dotseq - save by double` will crash Bespoke on loading.

NB: There seems to be a more generic issue with dotsequencer-snapshots combination, also with current nightly Bespoke versions . Some of my older savestates with a dotsequencer-snapshots combination, also have issue switching the dotsequencer snapshots: nothing seems to be changing. But at least these versions will not crash.
So perhaps this is related to the cause of the crashing of Bespoke double.


[snapshots dotseq.zip](https://github.com/user-attachments/files/19494358/snapshots.dotseq.zip)

NoxyNixie:
> [doubling-audio distortion.zip](https://github.com/user-attachments/files/19492783/doubling-audio.distortion.zip)

Thank you, this should be fixed now.

afentoe:
> > [doubling-audio distortion.zip](https://github.com/user-attachments/files/19492783/doubling-audio.distortion.zip)
> 
> Thank you, this should be fixed now.

Yes I can confirm, this is fixed now.

NoxyNixie:
Alright I think this is in a state again to do more testing.

I've applied the different approach to converting the loading data all across the board.
I've rolled back some of the doubling closer to where the audio buffer is manipulated to hopefully have less of a performance impact.
I've fixed a threading bug that I kept running into when testing.
Two saving bugs were fixed by @awwbees .
And I've merged in all the changes on the `main` branch.

Issue #1604, opened 2024-06-29
bug - multitrackrecorder crash the process if click "clear" while recording


johnny-smitherson:
## Describe the bug

Window crashes if you click on multitrackrecorder's "clear" button while it's still recording.

## Build info

bespoke 1.2.1 (Sep 16 2023 20:11:06) - Windows 11 - HEAD 9a2072e

## Steps To Reproduce

- add multitrackrecorder that receives audio (or open attachment)
- click "record"
- click "clear" whithout canceling record
- crash

## Expected vs. actual behavior

- expected recording to be stopped and track audio buffer cleared
- actually got window crash

## Attachment
[kickdrum-noise-reverb-for-line-clear.zip](https://github.com/user-attachments/files/16040139/kickdrum-noise-reverb-for-line-clear.zip)

## Workaround

- don't click clear without stopping record first

Issue #198, opened 2021-09-19
crash when clearing a running multitrack recorder


ambyshframber:
turning on record on a multitrack recorder, then pressing clear without first stopping the recording crashes the program. it doesn't happen every single time, but i managed to make it happen on demand a few times
[crash_2021-09-19_15-46.txt](https://github.com/awwbees/BespokeSynth/files/7192305/crash_2021-09-19_15-46.txt)

PR #1653, opened 2024-10-09
Force a poll to properly rebuild the buffers when the `multitrackrecorder` is cleared. Resolves: #1604.

NoxyNixie:
Force a poll to properly rebuild the buffers when the `multitrackrecorder` is cleared. Resolves: #1604.

There might be a better way but "this works :tm:". 

Maybe the code in Poll should be in Process, perhaps with a conditional bool that is set in the Clear method? Poke me if you want me to take that approach instead ;)


NoxyNixie:
Also resolves: #198.

PR #1674, opened 2024-10-23
Add `pulselimit` module

merkys:
`pulselimit` module limits the number of pulses it passes through it. It does so by keeping an internal counter of passed pulses, and when the number surpasses the user-entered number, it withholds the incoming pulses. There is also a reset button to reset the internal counter back to 0.


Source/PulseLimit.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+PulseLimit::~PulseLimit()
+{
+}
+
+void PulseLimit::CreateUIControls()
+{
+   IDrawableModule::CreateUIControls();
+
+   IDrawableModule::CreateUIControls();
```
NoxyNixie: This causes an assert to trigger. Only call this once!

merkys: Oh, that's an overlook. Thanks for spotting!

merkys:
Thanks a lot @NoxyNixie!

PR #1677, opened 2024-10-28
Add `taptempo` module

### LINKED ISSUE #1206: Tap tempo
## Summary

Adding a _tap tempo_ button to the transport module.

### Motivation

The feature would be a convenient way to set a tempo without thinking about it in terms of BPM, and without external MIDI hardware/software that supports tap tempo.

Using Bespoke's flexible signal routing system, the tap tempo button could in turn be connected to a MIDI button.

merkys:
This modulator module accepts notes and pulses and derives an average BPM based on incoming signals (a.k.a. tap tempo). I personally find tap tempo feature useful, as I do not know tempos by heart and sometimes enjoy changing them during performances.

Currently the module records the initial and the last note/pulse and divides the number of intermediate notes/pulses by the interval. There is a "Reset" button to clear the module.

Prior to merging, I would love to hear feedback on the following points:
1. <s>Name. Is `averagebpm` intuitive, or should it better be `taptempo`, or maybe something else?</s>
2. Calculation method. I have heard that some tools consider the last four-ish taps, while this PR considers all incoming.
3. <s>Clicking "Reset" button emits BPM of 0, which is probably unwanted. However I cannot find a way to prevent 0 from being transmitted (I have to return something, right?). Maybe a solution would be to store the last BPM in a variable and keep it unchanged until 2 notes/pulses are received.</s>
4. Saving/loading. Currently module does not recall anything. It should probably recall the last calculated BPM, but it seems that after loading it does not emit the 0 value, thus does not overwrite the setting of a connected module.

Fixes #1206


Source/AverageBPM.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+void AverageBPM::PlayNote(double time, int pitch, int velocity, int voiceIdx, ModulationParameters modulation)
+{
+   if (mEnabled && velocity > 0)
+   {
+      mLastBeatTime = time;
+      if (!mCount)
+		 mFirstBeatTime = time;
+	  mCount++;
+   }
```
PowerUser64: If you look closely, you're mixing tabs and spaces here (and in other places). For reference, Bespoke's source files are usually 3-space indented. 

There's an [editorconfig](https://editorconfig.org) file in the root of the repository that your editor should pickup if it supports it:
https://github.com/BespokeSynth/BespokeSynth/blob/840893a9fadf567cb96718311ae9ba7fdc1c1f99/.editorconfig#L5-L8

Source/AverageBPM.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+float AverageBPM::Value(int samplesIn)
+{
+   if (mCount < 2)
+	  return 0;
+
+   return (mCount - 1) / ((mLastBeatTime - mFirstBeatTime) / 60 / 1000) * 4;
```
PowerUser64: Have you considered making this a moving average? It would be make it easier to control in different situations, since you might want a certain amount of smoothing when triggering it by hand, but you might want less or no smoothing when triggering it with modulation.

EDIT: I misread this line, but my original question still stands.

PowerUser64 (changes requested):

Source/AverageBPM.cpp
merkys: You are right, I wrote this PR using different editors, one of them must have introduced tabs. I converted them to spaces.

Source/AverageBPM.cpp
merkys: Good idea. Should I fix the size of averaging window (say to 4), or put a slider for the user to choose it? Tapping after a long delay may still cause funny averages, but so does my current implementation.

merkys:
On Discord @NoxyNixie suggested renaming the module to `taptempo`. I like that. I will rename the module once @PowerUser64 comments are resolved, otherwise GitHub might lose them (not sure it properly tracks file moves).

Source/AverageBPM.cpp
PowerUser64: I would say probably allow a variable length with an integer slider, since it would make for more ways to use it.

I think the problem of what happens when there's a huge delay between pulses could be solved by keeping the reset button there. If it's being triggered by hand the user would just have to reset it before inputting the first pulse. That way, it can still be used for unusal purposes with modulation sources.

Source/TapTempo.h
```diff
@@ -0,0 +1,90 @@
...
+      width = mWidth;
+      height = mHeight;
+   }
+
+   TextEntry* mWindowEntry{ nullptr };
+   ClickButton* mReset{ nullptr };
+
+   int mCount{ 0 };
+   int mWindow{ 4 };
+   std::array<float, mWindow> mBeats;
```
NoxyNixie: I was expecting a std::vector here especially since you want it resizable (aka dynamic).

merkys: You are right. Although [std::vector.resize()](https://en.cppreference.com/w/cpp/container/vector/resize) does not do what I want out-of-the-box (when resized, does not end-align elements). But hey, I can store beats in reverse, i.e., store the most recent time as 0th element!

merkys: Fixed, marking as resolved.

merkys:
Thanks for formatting fix @NoxyNixie! I was about to address this on my next free time slot. The rest should be ready for review, module looks working to me<s>, but I have not checked whether all required parameters (`mWindow` only, basically) survive the save/load roundtrip</s>.

**Edit**: save/load roundtrip works as expected.

Source/AverageBPM.cpp
merkys: Rewrote the module according to @PowerUser64 comments, marking this thread as resolved.

Source/TapTempo.cpp
```diff
@@ -0,0 +1,128 @@
...
+   if (mCount > mWindow)
+      mCount = mWindow;
+}
+
+float TapTempo::Value(int samplesIn)
+{
+   if (mCount < mWindow)
+      return 0;
+
+   return (mCount - 1) / ((mBeats.front() - mBeats.back()) / 60 / 1000) * 4;
```
NoxyNixie: This `4` is the the number of beats in a bar?

If so then this should probably be `TheTransport->GetTimeSigTop()` instead. Or perhaps a combination of `TheTransport->GetTimeSigTop()` and `TheTransport->GetTimeSigBottom()`.

merkys: Right. There should be a better way to convert `time` (double) to seconds, yes?

NoxyNixie: Well not sure actually. Most of the timing code deals with measure length and subsets thereof.

There is `TheTransport->GetDuration(kInterval_1n)` but not sure you are looking for that.

merkys: OK, thanks, I will give it a look. I recall there is this `timerdisplay` module which knows time in seconds, maybe I could reuse what is done in it.

merkys: I see `timerdisplay` refers to double-valued global variable `gTime` which looks like it holds the global time in milliseconds.

PowerUser64: Why does the module need to know what the time signature is? In music, time signature is just how you interpret the tempo/bpm, no?

merkys: > Why does the module need to know what the time signature is? In music, time signature is just how you interpret the tempo/bpm, no?

Right, this should not be required in theory. What my module needs is time in seconds of every tap. This is independent from the time signature, in theory. Now I am not so sure exactly why multiplier of 4 is needed in this line, but without it I am getting incorrect results. I have a hunch that `time` value which is given to `PlayNote()` and `OnPulse` depends on beats in bar.

NoxyNixie: Maybe this `4` should be `mWindow`? I'm guessing here since I have not actually tested this code yet.

merkys: > Maybe this `4` should be `mWindow`? I'm guessing here since I have not actually tested this code yet.

No, the number of taps is accounted for by using `mCount` in the ratio. At this point `mWindow == mCount`.

PowerUser64: > I have a hunch that time value which is given to PlayNote() and OnPulse depends on beats in bar.

Have you tried using this taptempo module with other time signatures to see if it still works? I'd be interested to hear the results. I think I found something that seemed to indicate something was wrong with bespoke's interpretation of time signatures once, but I can't remember what it was.

merkys: > Have you tried using this taptempo module with other time signatures to see if it still works? I'd be interested to hear the results. I think I found something that seemed to indicate something was wrong with bespoke's interpretation of time signatures once, but I can't remember what it was.

You are right, changing time signature breaks the module, as it cannot correctly detect the bpm from 1n pulser pulses. Replacing `4` with `TheTransport->GetTimeSigTop()` solves the issue if the bottom stays constant at `4`, but I cannot quickly find how to factor the `TheTransport->GetTimeSigBottom()` in.

merkys: OK, I think I finally solved the contribution of both the top and bottom of the time signature in ef8d861a86c1799fa86fa03579150f2891955d55. The constant of `4` remains, but I cannot find an explanation of that. Now all time signatures result in the correct bpm as detected from 1n pulser.

awwbees:
I have a few issues from trying to test this:
- should this be a modulator at all? it seems like you'd only ever want to use this to set the tempo on the `transport` module, so maybe we should eliminate the step of needing to hook it up to the tempo slider, and instead just set it directly, and have this be an "other" module with no output
- it's cool that it can take a pulse, but it would probably make sense for it to have a "tap" button right on it, rather than needing to set up a `pulsebutton`
- am I supposed to tap every beat, or some other interval? it seemed like when I tapped every beat, I was getting very high tempos
- I was getting frequent crashes from the tempo getting set to 0, either from connecting it up to the tempo slider before tapping, or by pressing "reset". I'm not sure if I was using it in an unexpected way.
- from looking at the code, the calculation method seems overly simple and likely noisy (if the intended usage is for a human to tap it in), it's probably worth some research into seeing how other tap tempos do their calculations

NoxyNixie:
There is also #1703 which was closed without comment that had a different approach to tempo following. I attempted to open conversation with this person with a question but had no response (aside from the closing of the PR).

I do have concerns with #1703 though as it looks like code was derived from some place I couldn't verify (potential copyright or license issue). But the initial conversation didn't start so I've not mentioned that in that PR.

merkys:
Thanks for a review @awwbees!

> I have a few issues from trying to test this:
> 
> * should this be a modulator at all? it seems like you'd only ever want to use this to set the tempo on the `transport` module, so maybe we should eliminate the step of needing to hook it up to the tempo slider, and instead just set it directly, and have this be an "other" module with no output

Right, in most cases it should go directly to the `transport`, but sometimes a user (I) may want to just know the new tempo without setting it. Thus it could be hooked to some display module. Moreover, I can imagine wanting to have some part of my patch to follow other tempo than the main one (a quite eerie arrangement). I like BespokeSynth's modular approach, thus I think this could stay separate from `transport` to allow for some out-of-the-box usage.

> * it's cool that it can take a pulse, but it would probably make sense for it to have a "tap" button right on it, rather than needing to set up a `pulsebutton`

Agree, it could have a button of its own. Will do.

> * am I supposed to tap every beat, or some other interval? it seemed like when I tapped every beat, I was getting very high tempos

One should tap every beat. It is quite strange that you get very high tempos, this is a bug.

> * I was getting frequent crashes from the tempo getting set to 0, either from connecting it up to the tempo slider before tapping, or by pressing "reset". I'm not sure if I was using it in an unexpected way.

Right, I am aware of this issue, but could not come up with a solution yet.

> * from looking at the code, the calculation method seems overly simple and likely noisy (if the intended usage is for a human to tap it in), it's probably worth some research into seeing how other tap tempos do their calculations

Agree.

merkys:
I have added a "tap" button and solved the issue with "reset" button emitting 0 tempo value upon click. <s>What I would like to have as well, is the `taptempo` module not emitting 0 upon its initial patching to another module. I need to do some more reading to understand how to either override emitting, or read the target value upon patching.</s>

**Edit**: Prevented emitting value of 0 upon attaching to target when tempo is not yet calculated.

NoxyNixie:
`mTargetCable` was renamed everywhere to `mTargetCableSource`.

The change was part of https://github.com/BespokeSynth/BespokeSynth/commit/f742a57786bfe4100bd5a8c798e42e52c59d8c93

merkys:
Revisiting this PR and pooling all comments together, here is a to-do/to-discuss list:

- [ ] Calculation method. Now the durations between taps are plainly averaged. There might be better ways to do that. Maybe an addition of snapping mode (i.e., rounding the tempos to multiples of 10) would do.
- [ ] Saving/loading. Possibly the number of seen taps should be reset on load, otherwise the stored intervals might get crazily long.
- [ ] Should this be modulator or in-built transport feature? I personally vote for modulator to allow greater flexibility.
- [ ] Tap once per beat/Tap once per measure. Currently the latter is implemented, but now I think the former is more natural.

awwbees:
hi @merkys! coincidental timing, but I hadn't noticed your most recent post yesterday, and had semi-forgotten about this PR, but last night I suddenly had a desire for a tap tempo module so I built one. apologies for the toe-stepping! but I think at a high-level, this work has now been resolved, and any further iteration can be done on the new module.

merkys:
@awwbees - no worries! The important thing is that tap tempo is now implemented in BespokeSynth. Thanks!

Issue #1518, opened 2024-01-10
Architectural Problem - midi controller names aren't consistent across platforms


PowerUser64:
MIDI controllers do not have consistent names on all platforms. This is problematic for bespoke because controller json configurations rely on the name of the device to know which configuration to auto-load, so configurations will only auto load if they are loaded on the correct platform. I discovered this when I connected my launchpad X on windows and discovered that its name was different from what it was on linux.

| Linux | Windows |
| --- | --- |
| ![image](https://github.com/BespokeSynth/BespokeSynth/assets/24578572/2dcae684-a6b1-40d1-9d7a-de0d73d5be6a) | ![image](https://github.com/BespokeSynth/BespokeSynth/assets/24578572/2ae5dfcd-4e08-4cb0-bac3-500d508b24e8) |

To resolve this, I propose that something like a `names` array is added to the midi controller json files that contains an array of midi controller names. This can then be used to lookup device names and more accurately match them to layouts. Additionally, it might let us have more sensible names for some of the json files.

Build: f82acc4f2fae30d5f2d85fdb3c90dd71e928f55b

PR #1678, opened 2024-10-29
Enhanced midicontroller deviceOutName devicename matching In and Out

afentoe:
The `midicontroller` deviceout is generated from the devicein name by replacing the word Input to Output.

My MIDI devicename is "MIDISPORT 4x4 In A", "MIDISPORT 4x4 In B", etc... with matching output names like "MIDISPORT 4x4 Out A".

My deviceout would always be empty, even when manually selecting the correct device in the triangle/properties screen from the deviceout pulldown menu.

I have created a small fix and added a replace command to search for " In " and replace it with " Out ".
I have added spaces to avoid matching parts of words containing the string "In".

It is working now with my MIDI devices.

Possible future enhancement:
A better solution would be to keep automapping the output device, but when the user manually selects a deviceout from the triangle/properties screen, that selection should be kept, until a new midi input is selected in the main screen.

### DIFF
--- a/Source/MidiController.cpp
+++ b/Source/MidiController.cpp
@@ -2287,7 +2287,7 @@ void MidiController::ConnectDevice()
    ResyncControllerState();
 
    std::string deviceInName = mControllerList->GetLabel(mControllerIndex);
-   std::string deviceOutName = String(deviceInName).replace("Input", "Output").replace("input", "output").toStdString();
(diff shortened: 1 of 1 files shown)


PowerUser64:
Related: #1518

NoxyNixie:
Perhaps we need to figure out a better way of handling this but for the time being this should work.

afentoe:
Thanks for the merge.

Yes the split functionality between `midicontroller` and `midioutput`, and python `script` needing the `midicontroller` to send data, is all a bit confusing and would need a lot of rework.
But for now, 'it just works' 😃

PR #1681, opened 2024-10-30
Removed a limit of 10 functional modulation cables, you can now connect as many as your system will allow. Resolves: #1322.

### LINKED ISSUE #1322: Feat - allow macros to have more than 10 outputs or limit number of cables
## Summary

Currently, macrosliders and other modulators are limited to a total of 10 outputs. This limit is not communicated to users in any way. I think it would be good to communicate this limit to users by not letting them add more than 10 cables going out from a single patch point or allow a single modulation output to have more than 10 destinations.

In my opinion, it would be best to allow modulators to have more than 10 outputs, as the limit of 10 seems somewhat arbitrary. Otherwise, the number of cables should be limited to avoid confusion as to what is happening.

### Motivation

This would prevent confusion to users who may be wondering why their modulators only modify the first 10 things they are connected to.

NoxyNixie:
Removed a limit of 10 functional modulation cables, you can now connect as many as your system will allow. Resolves: #1322.

### DIFF
--- a/Source/FloatSliderLFOControl.cpp
+++ b/Source/FloatSliderLFOControl.cpp
@@ -254,8 +254,13 @@ void FloatSliderLFOControl::SetOwner(FloatSlider* owner)
    if (owner != nullptr)
       owner->SetLFO(this);
 
-   mTargets[0].mSliderTarget = owner;
-   mTargets[0].mUIControlTarget = owner;
+   if (mTargets.empty())
+      mTargets.emplace_back(owner, owner);
+   else
+   {
+      mTargets[0].mSliderTarget = owner;
+      mTargets[0].mUIControlTarget = owner;
+   }
 
    if (GetSliderTarget() != nullptr)
       InitializeRange(GetSliderTarget()->GetValue(), GetSliderTarget()->GetMin(), GetSliderTarget()->GetMax(), GetSliderTarget()->GetMode());
--- a/Source/FloatSliderLFOControl.h
+++ b/Source/FloatSliderLFOControl.h
@@ -73,7 +73,7 @@ public:
    void SetRate(NoteInterval rate);
    void UpdateFromSettings();
    void SetOwner(FloatSlider* owner);
-   FloatSlider* GetOwner() { return mTargets[0].mSliderTarget; }
+   FloatSlider* GetOwner() { return !mTargets.empty() ? mTargets[0].mSliderTarget : nullptr; }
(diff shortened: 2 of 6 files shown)


NoxyNixie:
I believe this change is also faster with less than 10 cables (which is very often) though I have not benchmarked this. Obviously with more than 10 I expect it to be slower than the previous version however it will have the correct behavior.

awwbees:
in a perfect world I'd profile this to verify how much it will affect performance, but my instinct is that adding additional performance tax to all usage of modulators is not worth improving the rare case that someone wants 10+ modulators. I would instead fix #1322 to limit the number of cables a user is able to split, and force the usage of a `macroslider` if someone wants more splits from a single modulation source.

NoxyNixie:
I don't agree. 

I also think that "additional performance tax" doesn't exist and I think with this patch the performance is better in general usage. 

The only place where performance might be slower is when new modulation cables are connected (due to resizing). In almost all other cases the iterations go from a static 10 to less than that so it would surprise me if that would cause "performance tax".

Ofcourse I can't say this for sure as I haven't measured this (yet, I do plan on testing this but currently lack time).

awwbees:
okay, I did some profiling on this, and determined that the performance is approximately equal before and after this PR. 

when doing the test though, I remembered my original motivation for using a statically-defined array rather than a resizable vector: avoiding thread safety issues, since the UI thread and the audio thread are both accessing that list. so, for example, you could get a crash in `GetSliderTarget()` if you have a modulator with 1 target, but the UI thread removes that target between the audio thread calling `!mTargets.empty()` and `mTargets[0].mSliderTarget`. this is a very tricky crash to actually reproduce, though, it requires precise unlucky timing. I'll try to find a way to hack things up to make it reliably crash, and then I have some ideas for how to avoid the issue.

the obvious way to avoid a crash here would be to add a mutex for access to mTargets, but that would introduce a big performance hit from the threads always waiting on each other.

awwbees:
here we go, the thread safety crash can be reproduced pretty quickly if you add a sleep here in `IModulator.h`:
```c++
// at top of file
#include "unistd.h"

   // replace GetSliderTarget() with this:
   FloatSlider* GetSliderTarget() const
   {
      if (!mTargets.empty())
      {
         usleep(1000);
         return mTargets[0].mSliderTarget;
      }
      else
      {
         return nullptr;
      }
   }
```

and create the setup below, and then remove the `envelope`'s modulator cable from the `gain`. should crash right away.
image

PR #1688, opened 2024-11-01
Update the Mathematical Expression Toolkit Library or "exprtk" to the latest version. Resolves: #1687.

### LINKED ISSUE #1687: Build breaks with clang-19: a template argument list is expected after a name prefixed by the template keyword [-Wmissing-template-arg-list-after-template-kw]
## Describe the bug

```
/wrkdirs/usr/ports/audio/bespokesynth/work/BespokeSynth-1.2.1/libs/exprtk/include/exprtk/exprtk.hpp:6408:49: error: a template argument list is expected after a name prefixed by the template keyword [-Wmissing-template-arg-list-after-template-kw]
 6408 |             expression_node<T>::ndb_t::template collect(branch_, node_delete_list);
      |                                                 ^
/wrkdirs/usr/ports/audio/bespokesynth/work/BespokeSynth-1.2.1/libs/exprtk/include/exprtk/exprtk.hpp:6471:49: error: a template argument list is expected after a name prefixed by the template keyword [-Wmissing-template-arg-list-after-template-kw]
 6471 |             expression_node<T>::ndb_t::template collect(branch_, node_delete_list);
      |                                                 ^
/wrkdirs/usr/ports/audio/bespokesynth/work/BespokeSynth-1.2.1/libs/exprtk/include/exprtk/exprtk.hpp:6515:49: error: a template argument list is expected after a name prefixed by the template keyword [-Wmissing-template-arg-list-after-template-kw]
 6515 |             expression_node<T>::ndb_t::template collect(branch_, node_delete_list);
      |                                                 ^
```

[log](https://pkg-status.freebsd.org/beefy18/data/main-amd64-default/p332c73b035e3_sf377a0c7dfa/logs/bespokesynth-1.2.1_1.log)

## Build info

Version: 1.2.1
Clang-19
FreeBSD 14.1

NoxyNixie:
Update the Mathematical Expression Toolkit Library or "exprtk" to the latest version. Resolves: #1687.

### DIFF
--- a/libs/exprtk/include/exprtk/exprtk.hpp
+++ b/libs/exprtk/include/exprtk/exprtk.hpp
@@ -2384,6 +2384,19 @@ namespace exprtk
             }
          }
 
+         static inline std::string seperator_to_str(const token_type t)
+         {
+            switch (t)
+            {
+               case e_comma : return ",";
+               case e_colon : return ":";
+               case e_eof   : return ";";
+               default      : return "UNKNOWN";
+            }
+
+            return "UNKNOWN";
+         }
+
          inline bool is_error() const
          {
             return (
@@ -3141,7 +3154,7 @@ namespace exprtk
 
                                  if (!operator()(t0))
                                  {
-                                    return i;
+                                    return 0;
                                  }
                               }
                               break;
@@ -3153,7 +3166,7 @@ namespace exprtk
 
                                  if (!operator()(t0, t1))
                                  {
-                                    return i;
+                                    return 0;
                                  }
                               }
                               break;
@@ -3166,7 +3179,7 @@ namespace exprtk
 
                                  if (!operator()(t0, t1, t2))
                                  {
-                                    return i;
+                                    return 0;
                                  }
                               }
                               break;
@@ -3180,7 +3193,7 @@ namespace exprtk
 
                                  if (!operator()(t0, t1, t2, t3))
                                  {
-                                    return i;
+                                    return 0;
                                  }
                               }
                               break;
@@ -3188,7 +3201,7 @@ namespace exprtk
                }
             }
 
-            return (g.token_list_.size() - stride_ + 1);
+            return 0;
          }
 
          virtual bool operator() (const token&)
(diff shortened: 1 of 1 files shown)


ArashPartow:
@NoxyNixie thanks for the help on this, but when you have a moment can you please close this PR..   👍 

I'm happy to do the update when the library has been tested and verified just like I did last year: https://github.com/BespokeSynth/BespokeSynth/pull/1165

NoxyNixie:
I can close it but am curious why though? Isn't this a perfectly valid version that resolves the issue posted in #1687?

Is the version I'm updating to not the latest?

ArashPartow:
@NoxyNixie  There's still some changes going on.  👍 

I would ask these questions before creating this kind of PR:

1. Does BespokeSynth need this particular point release? can we wait or is it do or die?
2. Does BespokeSynth build binaries/packages using a compiler that would break without this point release?

NoxyNixie:
> @NoxyNixie There's still some changes going on. 👍

Alright, but isn't this always the case with software?

> I would ask these questions before creating this kind of PR:
> 
> 1. Does BespokeSynth need this particular point release? can we wait or is it do or die?

Need is such a strong word. No it doesn't need anything however there is the desire to resolve bugs and reported issues. Yes we can wait but that is very far from my mindset. I'd rather solve the things I know how to solve now than to wait for a potentially better solution later.

> 2. Does BespokeSynth build binaries/packages using a compiler that would break without this point release?

Technically yes because the users are "bespoke" and a user ran into issues. 

We are also working towards a release and I personally would rather see this resolved sooner rather than later especially since it may take another year if not longer before we do our next release.

I am not stopping you from creating a PR in parallel with your version of the same resolution and would rather merge in your PR than mine. But I do want to merge in this fix with reasonable urgency.

If it is a matter of pride that your name is used instead of mine for the commit I'm good with that too. In either case you can already create your own PR right now.

ArashPartow:
@NoxyNixie I'm happy for you to handle this PR and merge it when it's ready. I'm just asking to hold off a bit, there's a chance there  may be a tag update in the next few days.

wrt what is in the new tag: There's only one bug fix in this current tag which is related to increasing the precision of asinh/acosh/atanh the inverse hyperbolic functions in terms of the double type when the input value is very large, and I'm not sure anyone that would be using BespokeSynth would have much use for such functions.  You would have seen this when I created the PR as I usually post the release notes: https://www.partow.net/programming/exprtk/exprtk_release_notes_v0.0.3.txt


All the other changes are related to evaluation performance, which should not be a show stopper for a pending BespokeSynth release. 

So Let us all agree to be astute engineers first and foremost and not slaves to gamification or engagement/virtue signalling.  👍

NoxyNixie:
Ah yeah BespokeSynth won't be affected as the current version only uses float types. However after the next release this will be changed as seen in #1645.

I can hold of a couple of days even weeks but probably not going to wait till 2025.

PR #1703, opened 2024-11-17
tempo follower + tap tempo

tonyrewin:


NoxyNixie:
Are you aware a tempo module was already being developed here: #1677?

PR #1754, opened 2025-02-11
Enhance python script functions me.get() and me.set() to search modules in prefab

afentoe:
There is no way to dynamically reference a module within a prefab from a python script in that same prefab, except by using a static path like `prefab~notesequencer~enabled`.

This PR enhances the functions `me.get()` and `me.set()` to search for modules in a prefab first, if the python script is running inside that prefab.
If the module is not found in the prefab, the functions will continue searching the module on the main screen.

This way, when a group of modules and the related python script, are moved to a prefab, the script will continue to work and will reference the modules inside the prefab, without any changes in the script.

Also duplicating a prefab, or loading a prefab from a saved preset, will keep working and will reference the new prefab path/name automatically.

In case a script in a prefab, wants to modify a module on the main screen directly, and there is a module with the same name inside that prefab, it's possible to prefix the path with `#`.
For example: modify the notesequencer on the main screen from a script in a prefab, which also contains a notesequencer:
```
me.set('#notesequencer~enabled', 1)
```

Breaking change:
Previously, if a script inside a prefab would modify a notesequencer on the main screen, it would use `me.get('notesequencer~enabled')`.
With this PR, if the script inside a prefab is run and in the prefab is a notesequencer as well, the module inside the prefab will now be modified when using `me.get('notesequencer~enabled')`.

I think this impact is very low and only changes behavior if a script is run within a prefab, and there are modules in and outside the prefab with the same name.

Note:
Existing scripts in a prefab would require a full path to a module inside that prefab, so including the prefab name: `prefab~notesequencer~enabled`
This PR will not break these scripts. Although adding the prefix again would result in an invalid path at first, `prefab~prefab~notesequencer~enabled`,
but the fallback check of the original path name, will then match the module in the prefab.

Note 2:
The first commit of this PR added a new function `me.get_path_prefix()` to get the current prefix.
This would require a lot of modifications of a script, if it is moved to a prefab.
For example:
```
pathprefix = me.get_path_prefix()
noteseq_active = me.get(pathprefix + 'notesequencer~enabled')
```

I have removed this new function again, in favor of modifying the `me.get()` and `me.set()` functions.

Concluding:
I hope this PR is accepted as it allows referencing modules inside of a prefab. Also the conversion of a group of modules and script into a prefab, is much more convenient.

I hope the breaking impact is considered acceptable. Only in a very specific case, the behavior is modified, as explained above.

### DIFF
--- a/Source/ScriptModule_PythonInterface.i
+++ b/Source/ScriptModule_PythonInterface.i
@@ -213,6 +213,18 @@ PYBIND11_EMBEDDED_MODULE(scriptmodule, m)
          return 0.0f;
       })
       ///example: pulsewidth = me.get("oscillator~pulsewidth")
+      .def("get_path_prefix", [](ScriptModule& module)
+      {
+         std::string path = module.Path();
+         if (ofIsStringInString(path, "~"))
+         {
+            return path.substr(0, path.find('~') + 1);
+         }
+         else 
+         {
+            return std::string("");
+         }
+      })
       .def("adjust", [](ScriptModule& module, std::string path, float amount)
       {
          IUIControl* control = module.GetUIControl(path);
--- a/Source/ScriptModule_PythonInterface.i
+++ b/Source/ScriptModule_PythonInterface.i
@@ -194,9 +194,36 @@ PYBIND11_EMBEDDED_MODULE(scriptmodule, m)
       }, "pitch"_a, "velocity"_a, "pan"_a = 0, "output_index"_a = 0)
       .def("set", [](ScriptModule& module, std::string path, float value)
       {
-         IUIControl* control = module.GetUIControl(path);
-         if (control != nullptr)
+         IUIControl* control = nullptr;
+
+         // Check if path is not empty and the first character is '#'
+         if (!path.empty() && path[0] == '#') {
+            // Remove the '#' from the path
+            path = path.substr(1);
+         }
+         else
+         {
+            // check if running in prefab
+            std::string script_path = module.Path();
+            std::string prefix = "";
+            if (ofIsStringInString(script_path, "~"))
+               prefix = script_path.substr(0, script_path.find('~') + 1);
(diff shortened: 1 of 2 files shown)


afentoe:
Reverted to draft, as this is only a partial solution.
For instance the `midicontroller.get()` function would also need to be modified, as well as other `...get()` functions.
So perhaps it's better to change it back to the `me.get_path_prefix`. This way the prefix can be used in the path parameter for all the `get` and `set` functions.
Will investigate this further.

NoxyNixie:
I'm in favor of keeping `me.get_path_prefix` as well as modifying all the `.get` methods as described in the original post.

Owh and I'm not sure your code allows for **nested** prefabs. Yes I know they are currently not possible but they used to be possible (and thus technically still loadable from old savestates) and it was expressed that we wanted to get nested prefabs back in the future if we can resolve the bugs around them.

afentoe:
@NoxyNixie 
I have investigated this further, and found the root cause of the issue is within the function `ScriptModule::GetUIControl(std::string path)`

2 bugs there:
- If path contains "~" then assumes the path contains prefab path, but would also match `notesequencer~length`  and no prefab prefix would be added. This was the root cause of the `me.get` issue.
- `Path()` was used to generate a prefix, but this will return `prefab~script`, so this generates an invalid path. This code was never hit, because there is always a ~ in the path, to reference controls on a module.

Regarding referencing a module on the main screen, there already is handling of the special char $ for that in the existing code.

Regarding prefabs in prefabs.
I added some tests for 2x ~ chars, but will see if this will work with prefabs inside prefabs.
I think I will have to copy the Path() up to the last ~, so then `prefab~prefab2~script` will return `prefab~prefab2~` 

I will create a new PR and close this PR when the new one is created.

afentoe:
Replaced by https://github.com/BespokeSynth/BespokeSynth/pull/1756

PR #1756, opened 2025-02-13
Fix python script functions me.get() and me.set() to use correct module paths in a prefab script

afentoe:
The python script functions `me.get()` and `me.set()` reference the main screen module instead of the module within that same prefab.

I have investigated this further, and found the root cause of the issue is within the function ScriptModule::GetUIControl(std::string path)

2 bugs there:
- If path contains ~ the code assumes the path is a prefab path, but would also match `notesequencer~length` and no prefab prefix would be added. This was the root cause of the `me.get` issue.
- `Path()` was used to generate a prefix, but this will return `prefab~script`, so this generates an invalid path. This code was never hit, because there is always a ~ in the path, to reference controls on a module.

The fix now generates the correct path of the prefab, if script is run within a prefab. So references will be to modules within the prefab.

See attached savestate for proof of concept.
[me.get test.bsk.zip](https://github.com/user-attachments/files/18783916/me.get.test.bsk.zip)

Note: This replaces https://github.com/BespokeSynth/BespokeSynth/pull/1754 which will be removed.

### DIFF
--- a/Source/ScriptModule.cpp
+++ b/Source/ScriptModule.cpp
@@ -782,10 +782,27 @@ void ScriptModule::PrintText(std::string text)
 IUIControl* ScriptModule::GetUIControl(std::string path)
 {
    IUIControl* control;
-   if (ofIsStringInString(path, "~"))
-      control = TheSynth->FindUIControl(path);
-   else
-      control = TheSynth->FindUIControl(Path() + "~" + path);
+   std::string prefix = "";
+
+   if (path == "")
+      return nullptr;
+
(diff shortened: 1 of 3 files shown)


afentoe:
As requested  by @NoxyNixie I also added new python function `me.get_path_prefix()` to retrieve current path when running script in a prefab module.

This function will return an empty string when running in main screen, or return the path prefix, when run from a prefab script, for example: `prefab~`

Example
```
active = me.get(me.get_path_prefix() + 'notesequencer~enabled')
```

Issue #1765, opened 2025-03-01
oscoutput with addresses using a hyphen or underscore doesn't save connections correctly


dowdm:
## Describe the bug

When using an OSC address with a hyphen or underscore, the connections to the module aren't loaded correctly after saving

## Build info

bespoke 1.3.0 (Dec 22 2024 18:35:38) - Windows 11 - HEAD 1d0f042

## Steps To Reproduce

1. add oscoutput
2. add 8 smoothers
3. connect smoothers to slider0-slider7 with default names
4. save
5. close project, reopen project
6. rename slider0-slider7 with a hyphen between slider and digit (slider-0, slider-1, etc)
7. save project
8. close
9. reopen project

## Expected vs. actual behavior

Expected: connections were as originally placed
Actual: some connections are missing

fireworx:
yeah connections are not saved :-( please fix

PR #1766, opened 2025-03-02
fix for issue #1765: oscoutput with addresses using a hyphen or underscore

seanwevans:


NoxyNixie:
You may have forgotten to commit your modifications of OSCOutput.h.

seanwevans:
whoops thanks, fixed

NoxyNixie:
hhmmm you can't change the signature of `SetUpFromSaveData` since it is an inherited and overridden method from the base class `IDrawableModule`.

I think the best approach in your case is to move the code that depends on `moduleInfo` to `LoadLayout` as opposed to `SetUpFromSaveData`.

(Also try to make it pass all the CI checks else we won't merge)

PR #1767, opened 2025-03-07
Fixed buffer overflow in NoteTable::PlayColumn causing unexpected crashes

afentoe:
When the `notetable` grid contains pitches > 127, several buffer overflows occur in NoteTable::PlayColumn.

Both mQueuedPitches and mPitchPlayTimes have a kMaxLength of 128, so pitches > 127 were causing buffer overflows.

To reproduce, create this config. No notes are required in the notetable, only pitches > 128/130.
image

### DIFF
--- a/Source/NoteTable.cpp
+++ b/Source/NoteTable.cpp
@@ -344,7 +344,7 @@ void NoteTable::PlayColumn(NoteMessage note)
    if (note.velocity == 0)
    {
       mLastColumnPlayTime[column] = -1;
-      for (int i = 0; i < 128; ++i)
+      for (int i = 0; i < mLastColumnNoteOnPitches.size(); ++i)
       {
          if (mLastColumnNoteOnPitches[column][i])
          {
@@ -362,7 +362,7 @@ void NoteTable::PlayColumn(NoteMessage note)
          int outputPitch = RowToPitch(row);
 
          // don't play notes > 127, and also to avoid bufferoverflow for mQueuedPitches and mPitchPlayTimes below
-         if (outputPitch > 127)
+         if (outputPitch >= mPitchPlayTimes.size() || outputPitch >= mQueuedPitches.size())
             continue;
 
          if (mQueuedPitches[outputPitch])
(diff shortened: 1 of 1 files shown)


Source/NoteTable.cpp
```diff
@@ -361,6 +361,10 @@ void NoteTable::PlayColumn(NoteMessage note)
       {
          int outputPitch = RowToPitch(row);
 
+         // don't play notes > 127, and also to avoid bufferoverflow for mQueuedPitches and mPitchPlayTimes below
+         if (outputPitch > 127)
```
NoxyNixie: Shouldn't this be `mQueuedPitches.size()` instead? Hardcoding 127 seems wrong to me.

afentoe: ok yes, that's cleaner.
Was investigating the overflow of all the possible definitions in NoteTable.h, which are hardcoded to kMaxLength (=128) or just set to 128, but this makes sense

afentoe: So it will be something like:
`if (outputPitch >= mQueuedPitches.size())`
or even better:
`if (outputPitch >= mPitchPlayTimes.size() || outputPitch >= mQueuedPitches.size())`

... as these are the 2 vars which have a potential buffer overflow in the code to protect.

I also see this line in the PlayColumn function:
https://github.com/BespokeSynth/BespokeSynth/blob/9e7f57e4dd2180451a9ea568f76f925f5d02f8ea/Source/NoteTable.cpp#L347

Should this be changed to a check with .size() as well? So something like this:

`for (int i = 0; i < mLastColumnNoteOnPitches.size(); ++i)`

I could also change these 2 if-tests to kMaxLength, instead of .size(), so that way link it to the definition. kMaxLength is used at more places in Notetable.cpp

It looks a bit cleaner this way:
`for (int i = 0; i < kMaxLength; ++i)`
and
`if (outputPitch >= kMaxLength)`


Please let me know how you want these 2 if statements to be modified, so I can change it in a new commit.

NoxyNixie: I think I'm impartial to either of those two approaches so long as it isn't a hardcoded number.

afentoe: ok, changed it to use <avar>.size() checks

PR #1768, opened 2025-03-11
add module.set_focus() to python script module

afentoe:
This adds a new Python function `module.set_focus(zoom)` to the `script` module.

Focus/zoom logic inspired (read: blatantly copied) from the `OSC controller` bespoke/module/focus function.

zoom = 0: zoom module to full screen
zoom = 0.1 - 8: zoom to specified zoomlevel

Code example:
```
ns = module.get("notesequencer")  
if ns is not None:
   ns.set_focus(0)
else:
   me.output('module not found')
```

### DIFF
--- a/Source/ScriptModule_PythonInterface.i
+++ b/Source/ScriptModule_PythonInterface.i
@@ -824,6 +824,25 @@ PYBIND11_EMBEDDED_MODULE(module, m)
             float value = ofClamp(control->GetValue() + amount, min, max);
             control->SetValue(value, ScriptModule::sMostRecentRunTime);
          }
+      })
+      .def("set_focus", [](IDrawableModule& module, float zoom)
+      {
+         ofRectangle module_rect = module.GetRect();
+         if (zoom >= 0.1)
+            gDrawScale = ofClamp(zoom, 0.1, 8);
+         else if (fabs(zoom) < 0.1) // Close to 0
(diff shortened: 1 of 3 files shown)


PowerUser64:
I have not used python with bespoke too much, so forgive me if this is not how things usually work, but could it be better to separate this into two functions? Those being a `module.get_pos()` function, in addition to a `bespoke.set_focus()` function? It could provide some more flexibility to the interface to separate them. Again, I might be off-base with this since I do not use python too often in bespoke.

It may also be worth seeing if it's possible to implement this with the LocationZoomer, which you could probably get some cool lerping out of, like what the minimap does when you save and load views.

NoxyNixie:
> I have not used python with bespoke too much, so forgive me if this is not how things usually work, but could it be better to separate this into two functions? Those being a `module.get_pos()` function, in addition to a `bespoke.set_focus()` function? 

Not sure what you are pointing at here but this PR only does one thing so there is no way to separate it?

> It may also be worth seeing if it's possible to implement this with the LocationZoomer, which you could probably get some cool lerping out of, like what the minimap does when you save and load views.

Possibly but I found (when working on the equivalent OSC implementation) that lerping here was more a distraction (and motion sickness inducing) than helpful. Especially when you are rapidly switching between different modules this way.

PR #1780, opened 2025-03-29
Vst multi channel out

geoffeep:
Added output pins to the VST Plugin node that map to outputs within the VST. 
Tested mostly with Kontackt VSTs on Windows 10

(re-opened pull request after fixing some formating issues!)


Source/LabelDisplay.cpp
```diff
@@ -27,6 +27,8 @@
 #include "SynthGlobals.h"
 #include "ModularSynth.h"
 
+#include <cstring>
```
NoxyNixie: Huh? Why did you add this?

Source/VoiceSetter.h
```diff
@@ -58,6 +58,8 @@ class VoiceSetter : public NoteEffectBase, public IDrawableModule, public IIntSl
       height = mHeight;
    }
 
+   static const int kMaxJuceMidiChannels = 16;
```
NoxyNixie: Unused variable?

Source/LabelDisplay.cpp
geoffeep: Erm. Yeah not sure how I did that. Fixing.

Source/VoiceSetter.h
geoffeep: That was supposed to be in VSTPlugin.h and then used in VSTPlugin.cpp and I must have forgot to do that.
Will also rename it max "stereo" channels because there was some interest on Discord from folks wanting mono out support which I might look into.

Source/LabelDisplay.cpp
NoxyNixie: You missed this one ;)

geoffeep: ... I am the worst, thanks for the patience :)

PR #1782, opened 2025-04-04
Add `dopplershift` module

merkys:
This PR introduces `dopplershift` module to model the Doppler shift of sound. It takes the speeds (in m/s) of both sound source and receiver and outputs a pitch shift value which can then be plugged in `pitchshift` module. It looks like this:

DopplerShift

Speeds are towards one another, thus can be negative. There is a "reverse" button to reverse both speeds, which can be used to model the receiver passing by the source. Speed of sound in medium can be changed using triangle menu.

I am interested as well in opinions whether this module belongs in BespokeSynth. Its use is rather specific and it does not do anything what cannot be done using a bunch of modulators. Nevertheless it greatly simplifies Doppler effect modelling, and I had fun writing it.

An example savestate:
[example__doppler.bsk.zip](https://github.com/user-attachments/files/19598152/example__doppler.bsk.zip)


merkys:
@NoxyNixie am I right to assume that "additional work required" label was added because failing checks? It seems that their failures were transient - I pushed a trivial commit causing checks to rerun, now none of them fail.

merkys:
Thanks for merge @awwbees!

PR #617, opened 2021-11-14
Add Launchpad mk2 and Launch Control XL controllers and add some scales

spectromas:
I've added Launchpad mk2 and Launch Control XL to the controllers directory (I can't get midi feedback for the Launch Control, I've posted a bug report  #606) and added some (mainly) Japanese scales.

PR #1784, opened 2025-04-10
feat: add Launchpad X controller definition

SpotlightKid:
This is apparently the same mapping as the "Launchpad mk2" in PR #617, but I think it still warrants its own file, since it is a separate model.

The mapping is the one when the device is in "Programmer Mode", which needs to be enabled by sending some SysEx to the device, but this is out of the scope of this PR.

### DIFF
--- a/resource/userdata_original/controllers/Launchpad X.json
+++ b/resource/userdata_original/controllers/Launchpad X.json
@@ -1,0 +1,46 @@
+{
+  "groups" : [
+     {
+        "rows": 8,
+        "cols": 8,
+        "position" : [ 0, 35 ],
+        "dimensions" : [28, 28],
+        "spacing" : [30, 30],
+        "controls" : [
+          81, 82, 83, 84, 85, 86, 87, 88,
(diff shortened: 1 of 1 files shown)


PowerUser64 (approved):
I've been using a file that's almost identical to this one for a while. Works great.

SpotlightKid:
> The mapping is the one when the device is in "Programmer Mode", which needs to be enabled by sending some SysEx to the device

Actually, you can also enter Programmer Mode by shortly holding the "Session" button, then pressing the orange clip launch button the lower right corner.

dromer:
I actually thought we had this already included!

Been also using a file pretty much exactly like this since a long time :+1:

afentoe:
> I actually thought we had this already included!
> 
> Been also using a file pretty much exactly like this since a long time 👍

LOL me too, I've been using it for the Launchpad MK2 for a long time as well

afentoe:
So perhaps I should add another PR for the `Launchpad MK2` by just copying and renaming this one, so it's finally officially imported 😄

PR #1786, opened 2025-04-12
Little refactors

PowerUser64:
I noticed some things that could be tweaked to make some parts of the code base a little more robust or easier to read. Here are some adjustments from what I noticed.

### DIFF
--- a/Source/ModuleFactory.cpp
+++ b/Source/ModuleFactory.cpp
@@ -567,10 +567,11 @@ std::vector<ModuleFactory::Spawnable> ModuleFactory::GetSpawnableModules(ModuleC
       }
    }
 
+   // Put effect modules at the end of the Audio category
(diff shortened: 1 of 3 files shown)


Source/ModuleFactory.cpp
```diff
@@ -720,33 +721,33 @@ ModuleCategory ModuleFactory::GetModuleCategory(std::string typeName)
...
+         break;
+
+      case SpawnMethod::Prefab:
+         return kModuleCategory_Other;
+         break;
+
+      case SpawnMethod::Preset:
+         return mFactoryMap[spawnable.mPresetModuleType].mCategory;
+         break;
+   }
```
PowerUser64, line 750: This actually should've been part of a commit previous to this one (3d0ab273f5d46a16632792dd4bea324099c84e90). I think git changed the line endings or something.

PR #1788, opened 2025-04-12
Add `padsynth` module

merkys:
This PR adds `padsynth` implemented as described [here](https://zynaddsubfx.sourceforge.io/doc/PADsynth/PADsynth.htm). <s>I open it as a draft PR first as there are still some cargo code (leftovers from copying KarplusStrong), but I would happily hear your comments.</s>

padsynth module in action

<s>The resulting sound is still a bit rough (as opposed to samples presented upstream), but I lack the understanding of the IFFT input generation algorithm to debug it further. Nevertheless it is an interesting sound to play with.</s>

The instrument sounds best with large buffer sizes. I thought that maybe it is worth using larger internal buffer and sliding window (I called this concept undersampling, feel free to correct me here) and added `mUndersample` scaling factor to `PadSynthVoice.h`. However, large values are CPU-intensive, as IFFT has to be computed for whole buffer at once. There is still room for improvement using rolling buffers.

As said, I am open to comments!


PowerUser64:
Nice work! I tested out the module and I have a few questions:
- Have you considered whether `offset` could be made to work independent of the buffer size? Usually it's best when buffer size doesn't impact the sound so it doesn't change with sample rate and such.
- Would it make more sense for `detune` to be called something else? It doesn't really sound like a traditional oscillator detune control, so maybe another name would be more fitting. From reading the dsp and your comments, it looks like it has to do with the "fuzz" or "spread" of the harmonics. Maybe **.
- What is the `scale` parameter doing? The word "scale" might not be the best for a parameter name, since "scale" can mean different things depending on the context (musical scale, as in "C Major" vs mathematical scale, as in "scalar value").
- How is the `oversample` option in the triangle menu supposed to behave? It seems to leave a lot of zeros in the signal where the karplusstrong module doesn't.

merkys:
Thanks a lot for your review and feedback, @PowerUser64! I will respond in-line below:

> * Have you considered whether `offset` could be made to work independent of the buffer size? Usually it's best when buffer size doesn't impact the sound so it doesn't change with sample rate and such.

Good point about keeping `offset` independent from buffer size. I thought about making it an integer offset instead of buffer fraction, but was worried about changing bounds on each buffer size change. I now see other modules doing so, thus I suppose it would be OK for `padsynth` as well. By the way, for `padsynth` the sound depends a lot on buffer size anyway due to the limit on used harmonics.

> * Would it make more sense for `detune` to be called something else? It doesn't really sound like a traditional oscillator detune control, so maybe another name would be more fitting. From reading the dsp and your comments, it looks like it has to do with the "fuzz" or "spread" of the harmonics. Maybe **.

I agree, `spread` would be much more intuitive, will change.

> * What is the `scale` parameter doing? The word "scale" might not be the best for a parameter name, since "scale" can mean different things depending on the context (musical scale, as in "C Major" vs mathematical scale, as in "scalar value").

I agree that parameter names here could be more intuitive (and consistent throughout BespokeSynth). According to [PADsynth description](https://zynaddsubfx.sourceforge.io/doc/PADsynth/PADsynth.htm), `scale` (or `bwscale` in original description) is *how much the bandwidth of the harmonic increase according to it's frequency*. Would `bwscale` or `bwfactor` be a bit clearer?

> * How is the `oversample` option in the triangle menu supposed to behave? It seems to leave a lot of zeros in the signal where the karplusstrong module doesn't.

I have not tested the `oversample` parameter. I "inherited" it from `karplusstrong`. I will see what I can do, but most likely I will drop it.

afentoe:
Cool! This looks really promising.

At the moment I see it does not compile.
Will you update this PR to make it compile again?

merkys:
@afentoe right, I see build fails on Windows, most likely due to issues with `fabs()` call. I will see what I can do.

merkys:
In https://github.com/BespokeSynth/BespokeSynth/pull/1788/commits/85e528986a83f6dbb5ab9587371b55411e560f48 I allowed negative values of `detune` parameter (will rename it later) which provides lower and darker bell-like sounds.

I as well exposed the `undersample` in triangle menu to allow extending the internal buffer by factor of 2^`undersample`, but for the price of higher CPU usage. Beware, large values may crash Bespoke due to memory exhaustion. I do not have an idea yet how to prevent this as crashes depend on buffer and RAM sizes. Maybe it would be best to put a lower upper limit based on RAM size of "regular" computer nowadays.

The commit should make the Windows build pass as I got rid of the problematic `fabs()` call.

afentoe:
ok thanks for trying to fix this.

I'm no expert in C++ or BespokeSynth development, but here are some general tips I found useful while I developed the `euclideansequencer`:
- Use CONTROL-K followed by CONTROL-D (or the equivalent in your editor) to reformat your code, so it will pass the code-formatting checks when compiling.
- Use clamp/min-max checks, especially where sliders are the source for values. Although the sliders have configured min/max values, it's easy to change these boundaries in the GUI, using CONTROL+Click. See what happens if you change the min value to -1, or the max value to a way higher number.
Especially check var arrays with fixed/max dimensions, and check that the (override) max value of a slider will not cause a bufferoverflow.
- When testing, I used some extreme automation by attaching LFO's to all of the sliders, and set them to vary the slider values in a high frequency, so do some "monkey" testing.

Did you find the specific compile errors in the windows build? You can find them in the job logs.

merkys:
Thanks @afentoe for very useful tips for testing! I did follow them, uncovered and clamped two control variables having possibility to cause segfaults. I have resolved the remaining code formatting and Windows build issues on the way.

I like how `euclideansequencer` uses float sliders for apparently integer variables allowing LFOs to be attached to them. <s>I do not quite understand how this is achieved, but I certainly would like to use this feature for `harmonics` parameter.</s>

**Edit**: Made `harmonics` slider a float slider. Now LFOs can be easily attached to it.

merkys:
TODO:

- [x] understand what `oversampling` does and either use it correctly, or remove altogether
- [x] find a way to limit `undersampling` so that larger values do not cause segmentation faults
- [x] allow LFOs to be attached to `harmonics` parameter
- [ ] better parameter names
- [ ] maybe decouple `offset` from buffer size
- [ ] think of how to avoid calling IFFT for each chunk of buffer when `undersampling` is > 1; maybe employ rolling buffer

merkys:
I allowed negative number of harmonics to be used. This switches from overtone to undertone series. I like the resulting darker sound.

afentoe:
Ah, yes, you found the reason👍.  I used float sliders because the int sliders do not allow LFO's to be attached. The easiest workaround was to use float sliders, but this introduced some issues, where rounding errors and comparison could cause issues.
For instance: 0.1 != 0.0, but when displaying the int value, it would always show as 0. So I had to add a lot of code for rounding/converting to int, to fix these issues.

The better solution would be to allow the LFO to also attach to a int sliders, but that was too complex for me at the time.

I will check out the Windows build. Great job so far.

merkys:
New stuff since the last time:
* Exposed a drop-down to change the method used to calculate amplitudes in the harmonics series. There is the original (which I named `step`), 1/sqrt(n) (named `sqrt`) and equal. Personally, I feel too little change between them than expected, thus this drop-down may go away soon.
* Found a way to limit the possible values for `undersample` to avoid segfaults.
* Fixed code formatting.

Sorry for editing git history with force pushes. I am experimenting with squashing as much as I can, at the same time working on a couple of computers, thus the resulting history is a mess. Maybe squashing was not the best idea under these circumstances.

merkys:
I found the reason for the rough sound: I had to separate real and imaginary parts for IFFT, instead of feeding it raw amplitudes and phases. Now this is fixed, and the instrument sounds much much nicer :relieved: And the amplitude method selection now properly affects the resulting sound.

I dropped the oversampling and slightly adjusted the parameter names.

I feel rather confident about this PR now, thus I mark it as ready for review. I would be grateful for reviews and comments.

awwbees:
finally taking a look at this, sorry for the extremely long delay. do you have any example .bsks or videos that demonstrate intended usage of this synth? I was able to drop it in and get some cool bizarre sounds out of it, but it felt pretty hard to be intentional with it, and I couldn't tell if what I was hearing was the intended sound or if there are bugs. I couldn't figure out how to achieve anything like the sounds in the [page you linked](https://zynaddsubfx.sourceforge.io/doc/PADsynth/PADsynth.htm)

merkys:
Thanks for a review! Padsynth sounds best with large sample sizes (as shown in the linked website). To achieve them, I use extended buffer sizes for `padsynth` module which could be achieved by increasing `undersampling` parameter in a triangle menu. With buffersize of 1024 and `undersampling` of 4 sounds made by `padsynth` resemble the ones in the linked website, at least to my ear. I will craft a .bsk or a video showing that.

Indeed, it is difficult to be intentional with this instrument, but I personally enjoy tinkering with it. Maybe some of the parameters could be hidden or moved to some "expert mode", but I made all of them visible just to experiment with this method of sound generation.

A big downside of the current implementation is that for each buffer filling it computes the extended buffer anew with inverse FFT, resulting in high CPU loads. Maybe the extended buffer could be stored and recomputed only when empty, but that might result in CPU load peaks.

PowerUser64:
I wonder if there's a way you could make this work independently of buffer size or sample rate. Maybe the module could have a buffer size of its own and ignore the global buffer size? It would have some "unusual" latency that way (since it'd act different from most (maybe all) other modules), but having a big project buffer adds lots of latency no matter what, so it might be the best of both worlds. Also, it would be nice to be able to change the buffer size or sample rate of bespoke and not change the sound of the patch a lot.

awwbees:
I agree, yes: adjusting buffer size in bespoke (or any DAW) should only affect performance and latency, and not have any impact on sound (the `feedback` module is an exception to this... but it's a pretty exceptional module by its nature). so it would be good to figure out a way to make this sound the same at all buffer sizes. I think it's too obscure of a thing to ask a user to change their buffer size to achieve a certain sound.

merkys:
Thanks for your input. I will rewrite the module to have a buffer of its own. I just have to figure out how to use a deque data structure in BespokeSynth.

PR #1791, opened 2025-04-28
make bespoke able to recover from an audio interface being disconnected and reconnected

awwbees:


PowerUser64:
Nice! Tested on Linux. With PipeWire, bespoke doesn't crash or stop working when the sound card is disconnected, even on current the main branch. I think the closest thing to an audio interface disconnect when using PipeWire is the sound backend is restarting (by running `systemctl --user restart 'pipewire*'` or by it crashing for some reason), which this PR doesn't address in its current state. This may or may not be worth looking in to fixing. The most popular DAWs on Linux (et. al. Ardour) can't recover from this, but it's also a state that doesn't happen very often.

To be clear, the "disconnected" state that I'm seeing is basically like the paused state from hitting the pause button. Selecting modules and moving them works, but the audio thread appears frozen (no sound plays, no notes or pulses are sent between nodes, no modulation updates, etc.).

PR #1792, opened 2025-04-30
Add FluidSynth module

mmlr:
This implements a native FluidSynth based module. A SoundFont can be specified and the presets/instruments for each channel can be selected.

The FluidSynth code is added as another submodule under libs but the build can be done in either of three ways:

* BESPOKE_FLUIDSYNTH=OFF to turn off the FluidSynth module entirely
* BESPOKE_SYSTEM_FLUIDSYNTH=ON to use a system provided library
* BESPOKE_SYSTEM_FLUIDSYNTH=OFF (default) to build from internal source

Using the system provided FluidSynth has the benefit of it possibly being configured to find a default SoundFont automatically so the module would work out of the box. Building from source on the other hand allows for a known version to be used (2.4.5 in this commit). When building from source, a stripped down library is built that excludes any audio drivers or MIDI IO as that is not needed here.


mmlr:
I'm currently working on this from the other end [here](https://github.com/FluidSynth/fluidsynth/pull/1564), by attempting to make it easier to embed FluidSynth via an abstraction that removes the need for the big GLib dependency. Once this is settled (either by being accepted or by switching to a fork with these changes) I will update this PR so that GLib isn't needed anymore. Pulling in such a big dependency feels inappropriate, seeing what it does to the CI pipeline alone.

NoxyNixie:
I noticed that this compiles now. Is it ready for review?

mmlr:
> I noticed that this compiles now. Is it ready for review?

Yes it is. On the FluidSynth side, the work to add an embedded OS abstraction has been merged and covers the use-case here, namely embedding FluidSynth into another project without pulling in additional dependencies. So this can now build just from a submodule like the other libraries that are already built this way in BespokeSynth.

ArkyonVeil:
Looking forward to seeing this happen. Cheers!

awwbees:
I'm finally taking a look at this, apologies for the extreme delay. one issue I'm finding with this from the development side is that the fluidsynth submodule includes a 425MB `test/` directory to download, which makes initializing submodules a prohibitively slow and costly experience, for no benefit (there's no reason for a bespoke developer to need that `test/` directory). I can't figure out a clean way to exclude that directory from being initialized, short of forking fluidsynth and deleting it. anyone have any ideas on how to avoid forcing a 425MB download on bespoke developers? that is a large enough barrier to block this from being merged, unfortunately.

awwbees:
separately, more on the topic of the content of this module: I can see how from a midi perspective, it makes sense to map channels to instruments. but in the bespoke paradigm, I don't think mapping voices to instruments makes sense like this. I think it would better match the bespoke paradigm to have a single dropdown for instrument selection, rather than one per voice.
image

PR #1793, opened 2025-05-03
Add angle controllers to rotate the circlesequencer rings

TristanCacqueray:
This change enables rotating the sequencer rings by a whole step.

### DIFF
--- a/Source/CircleSequencer.cpp
+++ b/Source/CircleSequencer.cpp
@@ -135,6 +135,12 @@ void CircleSequencer::DropdownUpdated(DropdownList* list, int oldVal, double tim
 {
 }
 
+void CircleSequencer::ButtonClicked(ClickButton* button, double time)
+{
+   for (int i = 0; i < mCircleSequencerRings.size(); ++i)
+      mCircleSequencerRings[i]->ButtonClicked(button);
+}
+
 void CircleSequencer::LoadLayout(const ofxJSONElement& moduleInfo)
 {
    mModuleSaveData.LoadString("target", moduleInfo);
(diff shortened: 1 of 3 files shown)


TristanCacqueray:
Hello, hope it's ok to make such a PR. First time user, thanks, BespokeSynth is amazing :)

merkys:
Thanks for a nice addition! I have tried it out, it works as expected, although I have not tested it under complex conditions. A pair of suggestions:

1. Add tooltips.
2. I think that having `<` and `>` buttons on both sides of `offset` slider might give the idea that the buttons are used to control the offset. I would suggest moving both buttons to one side of the slider.

TristanCacqueray:
Great, thank you for testing, I'll update the UI shortly!

merkys:
Thanks - the updated UI looks great! I like the placement of the arrows much better, and the tooltips make their functions clear. Some may argue on left/right vs. counter-clockwise/clockwise issue, but I like the tooltips short, and directions are intuitive to me.

TristanCacqueray:
You are welcome, thanks again for the feedback! To be honest, I came to BespokeSynth after someone shared a preset of my post on balanced rhythm ([link](https://midirus.com/zettle/music/rythmes)). I was presently surprised how easy it is to tweak the circular sequencer, and as a follow-up I'd like to add a dropdown list selector to enable loading one of the perfectly balanced rhythm (the list can be found at the end of my post link).

Please let me know if I missed something else for this PR, as far as I can tell it works as expected but I'm not super familiar with the system; I did try to connect a controller to the new button and was able to rotate the ring using another module.

PR #1805, opened 2025-05-24
Adjust modulation range when a single target is replaced

### LINKED ISSUE #1139: Bug - modulation range does not reset when changing a cable's destination
## Describe the bug

When changing a modulation cable's destination, the modulation range of the modulator doesn't reset in the same way it would if the cable was being initially dragged.

## Build info

b6e01b9e3d13654c6849f1389767b459e7c5d6a3

## Steps To Reproduce

[bespoke modulation range bug.webm](https://user-images.githubusercontent.com/24578572/216567865-e2981c5a-3736-4d7e-8184-b653125a7bc4.webm)

(notice how the values on the macroslider's range don't change when the cable's destination is changed)

The modulation range should get reset when the cable is connected to a new destination as if it was freshly connected. I don't have an example, but I have noticed this can be fairly problematic sometimes.

merkys:
This PR fixes #1139 by adjusting the modulation range not only for previously target-less modulator, but for single-targetted modulator whose single target gets replaced (previously there were no adjustment in this case).

The idea is to check the number of modulator's targets, detect the replacement of its single target and adjust the modulation range.

Since this PR touches `IModulator.cpp` it deserves a lot of testing to catch all possible corner cases. 

Fixes #1139

### DIFF
--- a/Source/IModulator.cpp
+++ b/Source/IModulator.cpp
@@ -41,7 +41,9 @@ IModulator::~IModulator()
 
 void IModulator::OnModulatorRepatch()
 {
-   bool wasEmpty = (mTargets[0].mUIControlTarget == nullptr);
+   int targetCount = 0;
+   for (size_t i = 0; i < mTargets.size() && mTargets[i].mSliderTarget != nullptr; i++)
+      targetCount++;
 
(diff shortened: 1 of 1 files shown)


merkys:
Thanks a lot for merging this @NoxyNixie!

PR #1807, opened 2025-05-25
Add Behringer BCR2000 controller definitions

stfn2xu:
BCR2000.json

	Matches the layout of the front pannel controls that emit MIDI
	plus two foot switch inputs, factory default first preset, first
	encoder group only.

BCR2000_encoder_groups.json

	The same except for having four copies of the top row of
	encoders, selectable with the four encoder group buttons on the
	hardware.

If you are not using multiple encoder groups then the "BCR2000.json" 
layout is a bit more compact and visually analogous to the hardware.

In both layouts the button directly above each knob in the first row of
knobs represents the push function of that encoder. The other three rows
of encoders have no such function in the hardware.

### DIFF
--- a/resource/userdata_original/controllers/BCR2000.json
+++ b/resource/userdata_original/controllers/BCR2000.json
@@ -1,0 +1,69 @@
+{
+   "groups": [
+      {
+         "rows": 1,
+         "cols": 2,
+         "position": [ 0, 0 ],
+         "dimensions": [28, 28],
+         "spacing": [35, 30],
+         "controls": [109, 110],
+         "messageType": "control",
+         "drawType": "button"
+      },
+      {
+         "rows": 1,
+         "cols": 8,
+         "position": [ 0, 50 ],
+         "dimensions": [28, 28],
+         "spacing": [30, 30],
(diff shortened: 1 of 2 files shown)


stfn2xu:
I see there is already a JSON file for the closely related BCF2000 and I tried to match the button/knob spacing of that approximately.  I don't have a BCF2000 on hand to compare functionality.

I tested these layouts with a single BCR2000 running the 1.10 firmware (latest/only AFAIK) and with both Bespoke Synth 1.2.1 (from... Ubuntu apt repositories?) and a local build on Ubuntu 24.04 of main at 0878b6c6 (to test the "no_grid" feature).

and in case anyone reviewing this change isn't familiar with the BCR2000...

- [product info](https://www.behringer.com/product.html?modelCode=0808-AAA)
- [manual](https://cdn.mediavalet.com/aunsw/musictribe/Mz1NG_3f5EKFj-TriSeXpQ/kD0JGO4iYkm1hTd4xWEHOQ/Original/BCF2000_BCR2000_M_EN.pdf) (via search on https://www.behringer.com/downloads.html)

PR #1809, opened 2025-05-29
Add `stereorotation` module

### LINKED ISSUE #1303: Feat - stereo rotation module
## Summary

Stereo [phase rotation](https://en.wikipedia.org/wiki/Rotation_matrix) is something that sometimes needs to be done to correct for phase issues. I built a prefab that can do this, but it uses a lot of CPU and it's pretty large. If this was done in a single module with a single parameter, that would be much preferable.

### Motivation

This would make it possible to fix complex stereo balance problems with only bespoke modules while being more efficient than if this is implemented in a prefab.

merkys:
This PR introduces `stereorotation` module as requested by @PowerUser64 in #1303. Given stereo input, it rotates the channels according to the given phase. Given mono input, it duplicates the input as the second channel and then performs the rotation. I am not 100% confident I am doing proper mono -> stereo conversion, thus it would be nice if someone more knowledgeable could give a look.

Fixes #1303


PowerUser64:
Sorry, didn't have time to test this! I'm sure it works great (math looks good). Can't wait to try it!!

merkys:
Great - let me know how it goes! If something does not work as intended, just ping me.

PR #1810, opened 2025-05-29
Fix stereo output for `seaofgrain` when mono sample is used (#1642)

### LINKED ISSUE #1642: seaofgrain only outputs the left channel when given a mono sample
## Describe the bug

The seaofgrain module only outputs the left channel when given a mono sample, regardless of the per-grain panning settings.

## Build info

bespoke 1.2.1 (Apr  4 2024 09:05:35) - Linux - makepkg 9a2072ed

## Steps To Reproduce

- Create a seaofgrain module and load a mono sample.
- Plug it into a new splitter, then connect both ends to the default gain.
- Note how only the left channel is present.

Both channels should be output, just like for stereo samples. Currently only the left one is present.

merkys:
This PR fixes #1642 (only one output channel is written by `seaofgrain` on mono samples) by using the same sample channel for both left and right for mono samples. It seems that previously all samples were treated as stereo samples, resulting in the right channel being outputted as empty for mono samples.

Fixes #1642

### DIFF
--- a/Source/SeaOfGrain.cpp
+++ b/Source/SeaOfGrain.cpp
@@ -261,6 +261,11 @@ float SeaOfGrain::GetSourceBufferOffset()
       return 0;
 }
 
+int SeaOfGrain::GetSampleNumChannels()
+{
+   return mSample->NumChannels();
+}
+
 void SeaOfGrain::FilesDropped(std::vector<std::string> files, int x, int y)
 {
    mLoading = true;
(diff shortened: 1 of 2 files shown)


merkys:
Thanks a lot for merging this PR in, @NoxyNixie! This issue has been troubling me for quite a while.

NoxyNixie:
You're welcome!

PR #1811, opened 2025-05-30
Clamp set values to each TextEntry range

stfn2xu:
Shenanigans that were permitted before this change include setting waveformviewer freq = 0 which crashed the GUI.


stfn2xu:
Manual test cases:

Screenshot from 2025-05-30 16-27-53

Prior to this change, both text entry would take on the connected `valuesetter` values and setting `freq` to 0 in `waveformviewer` would cause the GUI to become unresponsive while BespokeSynth consumes 100% CPU.

PR #1818, opened 2025-06-12
Make the `multitapdelay` UI more compact

merkys:
Currently the `multitapdelay` UI is not very compact: it lists tap sliders in one column, and with increasing the number of taps (currently by modifying variable `mNumTaps` in the code) new tap sliders appear below:
Screenshot from 2025-06-11 16-29-26

This PR places 4 taps on a single row with the same width as the buffer display:
Screenshot from 2025-06-11 16-32-20

Should the number of taps increase, new rows are added below, again with four taps on each of them:
Screenshot from 2025-06-11 16-33-38


PowerUser64:
NICEEE!!

PR #1821, opened 2025-06-13
Fix `stereorotation` output: audio output has to be added to the buffer, not written over it

merkys:
I made a subtle mistake in `stereorotation` module, causing it to overwrite its siblings' contributions to the target buffer. I should have used `+=` instead of `=`.

### DIFF
--- a/Source/StereoRotation.cpp
+++ b/Source/StereoRotation.cpp
@@ -67,13 +67,13 @@ void StereoRotation::Process(double time)
    {
       if (mEnabled)
       {
-         out->GetChannel(0)[i] = GetBuffer()->GetChannel(0)[i] * phaseCos - GetBuffer()->GetChannel(secondChannel)[i] * phaseSin;
-         out->GetChannel(1)[i] = GetBuffer()->GetChannel(0)[i] * phaseSin + GetBuffer()->GetChannel(secondChannel)[i] * phaseCos;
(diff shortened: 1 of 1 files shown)


PowerUser64:
Nice! I finally got a chance to check this out. I noticed the problem you fixed here, so good that you noticed it too.

One other thing I've noticed is that the parameters seem to only update once per buffer, which means that the slider always clicks (even with smoothing on it).

Also, I think it might make sense to make the range of the slider -0.5 to +0.5, since its center value is zero, and like the panner, it causes the opposite effect if you move it in the opposite direction. What do you think about this?

merkys:
> Nice! I finally got a chance to check this out. I noticed the problem you fixed here, so good that you noticed it too.

This is my first audio emitting module, thus a rookie mistake :smile:. I only recently found out how children audio buffers are written by their parents, this made me re-read my code.

> One other thing I've noticed is that the parameters seem to only update once per buffer, which means that the slider always clicks (even with smoothing on it).

Yeah, parameters are indeed not updated during the buffer processing run. I have seen `karplusstrong` doing live updates, should be easy to follow suit here.

> Also, I think it might make sense to make the range of the slider -0.5 to +0.5, since its center value is zero, and like the panner, it causes the opposite effect if you move it in the opposite direction. What do you think about this?

Yes, this makes total sense!

PR #1822, opened 2025-06-16
add mode to `notedelayer` to delay in milliseconds

awwbees:


PowerUser64:
YES! I kept making workarounds for this, but they were always bad. Thank you for this!

PR #1826, opened 2025-06-23
Add Python script commands for snapshots module

afentoe:
- Add some Python script commands to manage the `snapshots` module:
```      s.get_size()
      s.get_current_snapshot()
      s.has_snapshot(index)
      s.set_snapshot(index)
      s.store_snapshot(index, label)
         optional: label
      s.delete_snapshot(index)
```

- Minor changes to the `snapshots` module: input validation, add functions GetSize and SetLabel


Source/ScriptModule_PythonInterface.i
```diff
@@ -697,6 +698,50 @@ PYBIND11_EMBEDDED_MODULE(vstplugin, m)
...
+      .def("set_snapshot", [](Snapshots& snapshots, int index)
+      {
+         snapshots.SetSnapshot(index, gTime);
+      })
+      .def("store_snapshot", [](Snapshots& snapshots, int index, std::string label)
+      {
+         if (index >= 0 && index < snapshots.GetSize())
+         {
+            if (!label.empty())
+            {
```
NoxyNixie: Single lines within a scope do not need brackets and most of the Bespoke source code does not have brackets here (code style not caught by our formatting checks).

Source/Snapshots.cpp
```diff
@@ -307,9 +307,17 @@ void Snapshots::DrawModuleUnclipped()
 
 bool Snapshots::HasSnapshot(int index) const
 {
-   return !mSnapshotCollection[index].mSnapshots.empty();
+   if (index >= 0 && index < mSnapshotCollection.size())
+   {
```
NoxyNixie: Single lines within a scope do not need brackets and most of the Bespoke source code does not have brackets here (code style not caught by our formatting checks).

Source/Snapshots.cpp
```diff
@@ -307,9 +307,17 @@ void Snapshots::DrawModuleUnclipped()
 
 bool Snapshots::HasSnapshot(int index) const
 {
-   return !mSnapshotCollection[index].mSnapshots.empty();
+   if (index >= 0 && index < mSnapshotCollection.size())
+   {
+      return !mSnapshotCollection[index].mSnapshots.empty();
+   }
+   else
+   {
```
NoxyNixie: Single lines within a scope do not need brackets and most of the Bespoke source code does not have brackets here (code style not caught by our formatting checks).

afentoe:
@NoxyNixie OK I did a cleanup of the unneeded brackets of the reviewed lines.

I was a bit confused about the syntax, because in other parts of the source code, sometimes brackets are used surrounding single lines, for example:
https://github.com/BespokeSynth/BespokeSynth/blob/46f8b1f933562900944461085bd87aa0e6def821/Source/Snapshots.cpp#L123-L126
https://github.com/BespokeSynth/BespokeSynth/blob/46f8b1f933562900944461085bd87aa0e6def821/Source/ScriptModule_PythonInterface.i#L96-L101

afentoe:
Oh I will be checking the build pipelines over here, as my local build no longer works, because of the new Python checks.
I have Python 3.12.6, which now gives me this error:

    Could NOT find Python: Found unsuitable version "3.12.6", required range is
    "3.6...3.12" (found _Python_EXECUTABLE-NOTFOUNDfound components:

Although this Python range check is trying to avoid Python 3.13, it will also not allow 3.12.6.
I think by now, the Python issue was fixed, so perhaps this Python version check needs updating again.

Caused by this line now:
https://github.com/BespokeSynth/BespokeSynth/blob/46f8b1f933562900944461085bd87aa0e6def821/CMakeLists.txt#L80

NoxyNixie:
> I was a bit confused about the syntax, because in other parts of the source code, sometimes brackets are used surrounding single lines

Yeah there are bits and pieces all over the codebase that don't follow this style just yet but for the most part (and according to @awwbees ) we should use the style without brackets when possible.

> Oh I will be checking the build pipelines over here, as my local build no longer works, because of the new Python checks.

Yeah I'm trying out reverting that change now in a separate PR. Hopefully it works and we can switch back to the old way. Else we may have to do something fancy to essentially only delist 3.13.4 which was the affected version. 🤞

afentoe:
> Yeah I'm trying out reverting that change now in a separate PR. Hopefully it works and we can switch back to the old way. Else we may have to do something fancy to essentially only delist 3.12.4 which was the affected version. 🤞

It was _**3.13.4**_ specifically, but I think the cmake syntax for versions checks is not so advanced, to be able to exclude specific versions.
Good to see everything is working again, as the upstream issue was fixed in Python 3.13.5.

PR #1828, opened 2025-06-26
add colors to APC Mini and APC Key 25 to fix display of notesequencer grid

afentoe:
-Add colors to json of APC Mini and APC Key 25 midicontrollers to fix display of notesequencer grid.

This will show:
off: yellow (instead of 0 / off)
note: red
measurebar: green

NB1:
The APC does not understand NoteOff, so NoteOn velocity 0 is needed to be send, but this is not supported by the current midicontroller / grid. (This would require a call to SendNote with forceNoteOn)

NB2:
The color table for the APC Mini and the APC Key seems to be different, so different color tables are used.

-Rename APC MINI to uppercase to follow correct device name.
Layout is automatically loaded, but was displayed as 'default'. So by changing the filename to uppercase, it is now also displayed correctly as the active layout in the midicontroller.


afentoe:
I've added enhancement https://github.com/BespokeSynth/BespokeSynth/issues/1829 to create a final solution for this issue.

This PR is just a workaround by using the yellow color instead of velocity 0 for Off.

afentoe:
I'm waiting for testing / updating the colors array for supporting the radiosequencer with the APC Keys 25 by user `tonke` on Discord.
So please hold this PR until APC Keys 25 config is updated and confirmed.

afentoe:
Colors of the APC Keys 25 MK1 have been updated and tested as well.

So this PR is ready to be added.

NoxyNixie:
I'm not sure we should merge this as this works around or obscure what I consider a bug.

Velocity 0 noteon messages should work in my opinion and I'd rather see that fixed before re-evaluating this PR.

afentoe:
OK I understand your point of view.
I'm not sure if I can implement a fix for https://github.com/BespokeSynth/BespokeSynth/issues/1829 and add "ForceNoteOn" within the JSON template and handle that in the `midicontroller` module.

PR #1836, opened 2025-07-02
Set the titlebar to be dark themed on windows

jacksonmacc:
This PR changes the titlebar on Windows 11 (and some supported versions of 10 I'm pretty sure) to have a dark theme, to better match the rest of the application.

Before:
before

After:
after

This is my first time contributing to anything so please let me know if I've done anything wrong!

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -849,6 +849,7 @@ if (APPLE)
     endif()
 elseif (WIN32)
     target_compile_definitions(BespokeSynth PRIVATE BESPOKE_WINDOWS=1)
+    target_link_libraries(BespokeSynth PRIVATE "dwmapi.lib")
 
     IF (MINGW)
         target_link_libraries(BespokeSynth PRIVATE iphlpapi)
--- a/Source/Main.cpp
+++ b/Source/Main.cpp
@@ -165,29 +165,9 @@ public:
 #if BESPOKE_WINDOWS
          auto hwnd = getPeer()->getNativeHandle();
 
-         char buffer[4];
-         DWORD bufferSize = sizeof(buffer);
-         auto result = RegGetValueW(
-         HKEY_CURRENT_USER,
-         L"Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize",
-         L"AppsUseLightTheme",
-         RRF_RT_REG_DWORD,
-         nullptr,
-         buffer,
-         &bufferSize);
-
-         if (result == ERROR_SUCCESS)
-         {
-            // Windows registry values are in little endian
-            int i = int(buffer[3] << 24 |
-                        buffer[2] << 16 |
-                        buffer[1] << 8 |
-                        buffer[0]);
-
-            BOOL darkMode = i != 1;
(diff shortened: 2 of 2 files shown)


NoxyNixie:
Does this always force it dark mode?

If so I'd like to see it follow the windows setting instead. So when people have windows set to be dark mode it makes Bespoke dark mode. If windows is set light mode it sets or keeps the light mode for Bespoke.

jacksonmacc:
Ok, it matches user preferences now. The way its implemented is a bit weird, it involves directly checking windows registry values but I don't think there's a better way of doing it. I would love to be proven wrong though!

(The only alternative method I could find involved checking the user's set background colour and calculating "percieved brightness," so I went with what I thought was the lesser of the two evils because the registry method just gives a boolean value)

NoxyNixie:
I think accessing the registry is fine. I'll review this tomorrow.

NoxyNixie:
hhmmm odd, when it spawns i shows like so:

image

but then once I click on the window it goes to:

image

NoxyNixie:
This might have to do with me running it in Debug mode though (and yes my dark mode is a bit darker than yours but that shouldn't matter much).

jacksonmacc:
I just double-checked, and mine doesn't do that in either debug or release. Are you using windows 10?

NoxyNixie:
Yes, the superior one!

jacksonmacc:
Thats fair enough, honestly. The feature exists on 10 but it doesn't seem to be officially supported. [This](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/ne-dwmapi-dwmwindowattribute) says its only 11 and up (DWMWA_USE_IMMERSIVE_DARK_MODE), so I'll try checking the version and see if that fixes it.

NoxyNixie:
Well it kinda works, just need to click the window (or anywhere in the window) once and it is fine. Most people will click pretty quickly within Bespoke's window :P

jacksonmacc:
I realised I hadn't manually refreshed the window after changing the titlebar, so that could be the fix. If not, I think it might just be that the feature isn't fully supported.

NoxyNixie:
hhhmmm look what I stumbled on:

https://github.com/juce-framework/JUCE/commit/3d282c10784661789881a835779d76f606556465

We may want to update JUCE so that this can be done through its API and make it work for all platforms.

NoxyNixie:
You know what I didn't look at the date, that was in 2021. We should have JUCE API for doing darkmode already.

jacksonmacc:
Fixed, thanks for the catch! I don't think juce can actually change the theme itself though with a native titlebar, so we still have to drop down to win32 for that.

NoxyNixie:
I wonder if we can set this before the window is created?

jacksonmacc:
I can move the code into initialise, but in order to change the theme we need access to the HWND pointer and that doesn't get created until the window gets created.

NoxyNixie:
So there are no flags on creation?

jacksonmacc:
There is, but not for the titlebar theme. The usual way of doing it is create the window (which is hidden by default), change all settings and then refresh and show the window. I'm not sure how or if that translates to juce though.

PR #1864, opened 2025-07-23
Adjust buffersize in the looper to make it less likely to access beyond the buffer. Resolves: #1500.

### LINKED ISSUE #1500: Bug - looper occasionally causes bespoke to crash when set to 16 bar length
## Describe the bug

When set to length 16, the looper can cause bespoke to crash. I have discussed this with a few others on discord and we all agree that it's a reoccurring problem.

## Build info

f82acc4f2fae30d5f2d85fdb3c90dd71e928f55b

## Steps To Reproduce

(inconsistant)

image

1. open a blank project
2. add a looper
3. set it to length 16
4. if you're lucky (or unlucky), it'll crash

Seems to happen about 20% of the time for me.

The looper should not cause bespoke to become unstable when set to any supported bar length.

NoxyNixie:
Adjust buffersize in the looper to make it less likely to access beyond the buffer. 

The only instance where this will still happen is if the tempo is lowered and then the number of bars is increased.
Adjusting the buffer sizes in realtime is likely not a good idea.

Resolves: #1500.

### DIFF
--- a/Source/Looper.cpp
+++ b/Source/Looper.cpp
@@ -50,9 +50,14 @@ Looper::Looper()
 : IAudioProcessor(gBufferSize)
 , mWorkBuffer(gBufferSize)
 {
-   //TODO(Ryan) buffer sizes
-   mBuffer = new ChannelBuffer(MAX_BUFFER_SIZE);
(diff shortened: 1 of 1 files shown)


NoxyNixie:
Other solution will be used.

PR #1892, opened 2025-09-04
Add `binaryvalue` module

merkys:
This PR introduces `binaryvalue` modulator which takes in a number in range [0, 255], converts it to binary and sends each bit to a separate modulation cable:

Screenshot from 2025-09-04 08-01-18

Apart from the standard binary representation, it supports [Gray code](https://en.wikipedia.org/wiki/Gray_code) which has a nice property that two neighboring numbers differ by exactly one bit, making value transitions very smooth.

The purpose of this module is to control 8 boolean modulation values (think enable/disable, button clicks, etc) by changing one number.


Source/ModulatorBinaryValue.h
```diff
@@ -0,0 +1,108 @@
...
+   };
+
+   BitModulator mBit0;
+   BitModulator mBit1;
+   BitModulator mBit2;
+   BitModulator mBit3;
+   BitModulator mBit4;
+   BitModulator mBit5;
+   BitModulator mBit6;
+   BitModulator mBit7;
```
NoxyNixie: I'm not so sure about this code. Whenever variables are used and are numbered in this way I can't help but think that this should be done differently. Also why do these need to be here in the first place? Isn't the array below enough? You'll have to change the way you initialize the array but that should be easy.

merkys: Fair point, I could have looked for a better solution. Will work on that.

Source/ModulatorBinaryValue.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+void ModulatorBinaryValue::DrawModule()
+{
+   if (Minimized() || IsVisible() == false)
+      return;
+
+   mInputSlider->Draw();
+   mCodeSelector->Draw();
+
+   for (size_t i = 0; i < 8; ++i)
```
NoxyNixie: Most of bespoke uses `int`.

Source/ModulatorBinaryValue.cpp
```diff
@@ -0,0 +1,128 @@
...
+   {
+      mBits[i].GetCableSource()->SetManualPosition(mWidth / 9 * (i + 1), mHeight);
+      DrawTextNormal(ofToString(GetBitValue(7 - i)), mWidth / 9 * (i + 1) - 4, mHeight - 9);
```
NoxyNixie: Mac build fails here due to implicit conversion of size_t to int which loses precision. I think using `int` in the loop will fix this.

Source/ModulatorBinaryValue.h
merkys: Fixed this by introducing an array and initializing bit values in modulator's constructor. At the same time I have introduced a constant declaring the number of bits used in the module, just to have it in a single place.

Source/ModulatorBinaryValue.cpp
merkys: Thanks, switched to `int` everywhere in the module.

Source/ModulatorBinaryValue.cpp
merkys: Replaced with `int`, so Mac build works now.

PowerUser64:
Very cool!

merkys:
Thanks @PowerUser64! :smile:

PR #1895, opened 2025-09-10
Implement "reset on load" mode for `pulselimit`

merkys:
This PR introduces `reset_on_load` boolean value in triangle menu of `pulselimit`, allowing to use it as "loadpulse". Enabling `reset_on_load` (off by default) makes it reset module's internal pulse counter to 0 every time a savestate containing it is loaded.

Creating as draft PR first as I need to test the backwards compatibility.


merkys:
Tested old savestates, all works well.

PR #1896, opened 2025-09-10
Increase the buffer in `spectrum` to avoid segfaults on larger buffersizes

### LINKED ISSUE #1837: `spectrum` overflows buffer with buffersize > 1024
Adding `spectrum` to a patch causes immediate segmentation fault with buffersize of 4096. Here is the `gdb` backtrace:
```gdb
(gdb) backtrace
#0  0x00007ffff6eb2577 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x0000555555f02036 in RollingBuffer::WriteChunk (this=this@entry=0x55555e7e36e8, samples=samples@entry=0x555556bf4520 <gWorkBuffer>, size=4096, channel=channel@entry=0) at ./Source/RollingBuffer.cpp:84
#2  0x0000555555fcbbac in SpectralDisplay::Process (this=0x55555e7e3080, time=<optimized out>) at ./Source/ChannelBuffer.h:48
#3  0x0000555555bc0c22 in ModularSynth::AudioOut (this=0x55555703aa78, output=0x7ffef1dc6aa0, bufferSize=4096, nChannels=16) at /usr/include/c++/14/bits/stl_vector.h:1128
#4  0x0000555556125c1a in juce::AudioDeviceManager::audioDeviceIOCallbackInt (this=0x55555703a710, inputChannelData=0x0, numInputChannels=0, outputChannelData=0x7ffef1dc6aa0, numOutputChannels=16, 
    numSamples=4096, context=...) at /usr/share/juce/modules/juce_audio_devices/audio_io/juce_AudioDeviceManager.cpp:1020
#5  0x000055555611e755 in juce::(anonymous namespace)::ALSAThread::run (this=0x7ffef1ef8190) at /usr/share/juce/modules/juce_core/memory/juce_HeapBlock.h:191
#6  0x00005555560d1d58 in juce::Thread::threadEntryPoint (this=0x7ffef1ef8190) at /usr/share/juce/modules/juce_core/threads/juce_Thread.cpp:110
#7  0x00005555560d1f1d in juce::juce_threadEntryPoint (userData=<optimized out>) at /usr/share/juce/modules/juce_core/threads/juce_Thread.cpp:132
#8  operator() (__closure=0x0, userData=<optimized out>) at /usr/share/juce/modules/juce_core/native/juce_Threads_linux.cpp:52
#9  _FUN () at /usr/share/juce/modules/juce_core/native/juce_Threads_linux.cpp:55
#10 0x00007ffff6de0b7b in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#11 0x00007ffff6e5e7b8 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
```
This happens since at least `v1.1.0` release.

merkys:
This PR addresses #1837 by increasing the number of FFT bins in `spectrum` to 1024 * 8 (the same as in `eq` module). Without this change, `spectrum` segfaults with buffersizes > 1024. I do not consider this PR to fix completely #1837 as a better fix would be to base the number of FFT bins on the buffersize, but I need more time to analyze the possible ramifications of doing so.


NoxyNixie:
I'm not sure I should merge this. This on its own has a significant impact on performance in all use cases.

`kNumFFTBins = 1024`:
image

`kNumFFTBins = 1024 * 8`:
image

Aside from the performance impact it also changes the visualization which is more accurate with more bins (as expected).

Changing the bin size to be the same as buffersize is also not a good idea since usually you'd expect better performance with larger buffersizes since the system can then catch up more.

Ideally the code should be changed so that it works with 1024 (or even changeable but the default would be 1024) bins on all buffersizes.

merkys:
Thanks for in-depth analysis of the change. Given that, I now lean to the idea of leaving 1024 bin size. In that case a solution would be to split the buffer into segments of 1024, summing them and averaging over time. I will give it a look to see how to implement this.

PR #1897, opened 2025-09-16
Add `velocitytoduration` module

merkys:
This PR introduces a `velocitytoduration` module which sets note's duration to its velocity divided by 127 and multiplied by maximum duration set via a slider.

Motivation: I am building a MIDI instrument with vibration motors, but their vibration speed (or intensity) cannot be made proportional to MIDI note velocity. A solution I see is to convert note's velocity to its duration.


Source/VelocityToDuration.cpp
```diff
@@ -0,0 +1,127 @@
...
+   }
+}
+
+void VelocityToDuration::FloatSliderUpdated(FloatSlider* slider, float oldVal, double time)
+{
+}
+
+void VelocityToDuration::LoadLayout(const ofxJSONElement& moduleInfo)
+{
+   mModuleSaveData.LoadString("target", moduleInfo);
```
NoxyNixie: This is no longer needed.

Source/VelocityToDuration.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+   SetUpFromSaveData();
+}
+
+void VelocityToDuration::SetUpFromSaveData()
+{
+   SetUpPatchCables(mModuleSaveData.GetString("target"));
```
NoxyNixie: This is no longer needed.

Source/VelocityToDuration.cpp
```diff
@@ -0,0 +1,127 @@
...
+   {
+      PlayNoteOutput(note);
+      return;
+   }
+
+   if (note.velocity > 0)
+   {
+      PlayNoteOutput(note);
+
+      ComputeSliders(0);
```
NoxyNixie, line 88: This is expensive, is this absolutely necessary? i.e. for realtime modulation? Would you modulate the sliders used in this module in the first place?

Source/VelocityToDuration.cpp
merkys: Removed, thanks!

Source/VelocityToDuration.cpp
merkys: Removed, thanks!

Source/VelocityToDuration.cpp
merkys: I took this code from `Source/NoteSustain.cpp`. I guess one might want to modulate the slider used here.

NoxyNixie: You sure? To me it seems like this is not something that would be good with modulation due to the nature of the module.

To me it seems that the result would be somewhat random and I feel that there are better ways doing things randomly.

merkys: I do not see why not: this parameter is essentially controlling the scale (multiplier) for converting velocity to duration, thus a user might want to change it. Surely there are other means to do so, for example, by using velocity scaler. I do not know for sure what `ComputeSliders(0)` does, but here it is called only on incoming notes. Or am I missing something?

NoxyNixie:
You should merge in the current head so that you have the pybind11 fix. That way the CI does not fail.

merkys:
Thanks a lot @NoxyNixie!

PR #1898, opened 2025-09-17
Bump `cmake_minimum_required` to 3.5

merkys:
Newer CMake versions have retired compatibility with versions < 3.5. Thus this now fails Mac builds in the pipelines. This PR bumps `cmake_minimum_required` to 3.5.


merkys:
Oh wait, the build failure was because `cmake_minimum_required` was too low in pybind11 submodule, not the main BespokeSynth, which already has 3.16! Sorry for the noise.

PR #1927, opened 2025-11-09
doc(taskfile): add build automation tools and instructions in README,…

davehorner:
… include ensure_task scripts and Taskfile.yml

similar to https://github.com/BespokeSynth/BespokeSynth/pull/1922

I plan to improve this more; possibly make the install actually copy to a public path instead of installing the prereqs.

I don't know if this needs to go upstream or not; i find it useful and thought I'd share.
--dave

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,3 +102,47 @@ To be able to build you will need a few things, depending on your OS
 * On Linux you probably already have everything (gcc, git, etc...), but you will need to install required packages. The full list we
   install on a fresh ubuntu 20 box are listed in the azure-pipelines.yml
     * Some distributions may have slightly different package names like for instance Debian bookworm: You need to replace `alsa` and `alsa-tools` with `alsa-utils`
+
+### Using Go Task for Build Automation
+
+BespokeSynth has a [Go Task](https://taskfile.dev) Taskfile.yml as a cross-platform build automation tool. This makes it easy to configure, build, and run BespokeSynth with a single command.
+
+```sh
+task           # Configure and build BespokeSynth (default)
+task run       # Run the built BespokeSynth executable
+task clean     # Remove build artifacts
+task install   # Install build prerequisites for your platform
+```
+
+Task will automatically detect your platform (Windows, macOS, Linux, or WSL) and use the appropriate build directory and commands.
+
+For more details, see the `Taskfile.yml` in the repository root.
+
+---
+
+### One-Step Build Scripts: ensure_task
+
+For convenience, you can use the provided scripts to automatically install Go Task (if needed) and build BespokeSynth in a single step:
+
+- **Windows:** `ensure_task.bat`
+- **Linux/macOS/WSL:** `ensure_task.sh`
+
+#### Usage
+
+**On Windows (cmd.exe):**
+
+```cmd
+ensure_task.bat [task arguments]
+```
+
+**On Linux/macOS/WSL (bash):**
+
+```sh
+./ensure_task.sh [task arguments]
+```
+
+- If Go Task is not found, the script will attempt to install it using your system's package manager or by downloading a release.
+- Any arguments you provide will be passed directly to `task` (e.g., `run`, `clean`, etc.).
+- By default, running the script with no arguments will perform the default build.
+
+See the script files for more details or to customize the installation logic for your environment.
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -1,0 +1,193 @@
+version: "3"
+
+vars:
+  BUILD_DIR: '{{default "ignore/build" .BUILD_DIR}}'
+  CONFIG: '{{default "Release" .CONFIG}}'
+  # For Windows MSVC generator:
+  VS_GEN: 'Visual Studio 17 2022'
+  VS_ARCH: 'x64'
+
+tasks:
+
+  default:
+    desc: Build BespokeSynth (configure + compile)
+    deps: [submodules]
+    cmds:
+      - task: configure
+      - task: build
+
+  submodules:
+    desc: Init / update git submodules recursively
+    cmds:
+      - git submodule update --init --recursive
+
+  # -------------------------------
+  # Install: meta helpers
+  # -------------------------------
+  install:
+    desc: Install build prerequisites (chooses a platform task)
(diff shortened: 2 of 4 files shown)


davehorner:
i'm going to work on this some more and I wonder if you'll accept the mix of technologies...

Taskfile.yml
```diff
@@ -0,0 +1,193 @@
...
+          sudo apt install -y libwebkit2gtk-4.1-dev
+          if [ -f /usr/lib/x86_64-linux-gnu/pkgconfig/webkit2gtk-4.1.pc ] && \
+             [ ! -f /usr/lib/x86_64-linux-gnu/pkgconfig/webkit2gtk-4.0.pc ]; then
+            sudo ln -s /usr/lib/x86_64-linux-gnu/pkgconfig/webkit2gtk-4.1.pc \
+                       /usr/lib/x86_64-linux-gnu/pkgconfig/webkit2gtk-4.0.pc
+          fi
+        else
+          # On some distros it may still be 4.0
+          sudo apt install -y libwebkit2gtk-4.0-dev || true
+        fi
```
PowerUser64, line 91: Same as an earlier comment, I don't think we should install things on the user's computer. Also, this assumes that the user is running ubuntu and makes changes to /usr, which is a bit sketchy.

Taskfile.yml
```diff
@@ -0,0 +1,193 @@
...
+    cmds:
+      - |
+        .\\ignore\\build\\Source\\BespokeSynth_artefacts\\Release\\BespokeSynth.exe
+
+  run:posix:
+    platforms: [linux, darwin]
+    cmds:
+      - |
+        BIN="{{.BUILD_DIR}}/Bespoke/BespokeSynth"
+        [ -x "$BIN" ] || BIN="{{.BUILD_DIR}}/BespokeSynth"
```
PowerUser64, line 187: I don't think this path is right. Wouldn't this expand to `ignore/build/Bespoke/BespokeSynth`? The bespoke executable is placed in the same spot that it is on windows, just without the .exe at the end.

Taskfile.yml
```diff
@@ -0,0 +1,193 @@
...
+    cmds:
+      - task: run:windows
+      - task: run:posix
+    silent: true
+
+  run:windows:
+    platforms: [windows]
+    cmds:
+      - |
+        .\\ignore\\build\\Source\\BespokeSynth_artefacts\\Release\\BespokeSynth.exe
```
PowerUser64, line 180: Shouldn't this use `{{.BUILD_DIR}}` like it does elsewhere?

Taskfile.yml
```diff
@@ -0,0 +1,193 @@
...
+    desc: Build with MSVC
+    platforms: [windows]
+    cmds:
+      - cmake --build "{{.BUILD_DIR}}" --config {{.CONFIG}} --parallel
+
+  build:posix:
+    desc: Build with Makefiles/Ninja on POSIX
+    platforms: [linux, darwin]
+    cmds:
+      - cmake --build "{{.BUILD_DIR}}" --parallel
```
PowerUser64, line 147: Should this include a `--config` flag, like the windows one has?

ensure_task.sh
PowerUser64, line 1: Same comment as on the other script, I don't think we should install things on the user's computer, it's not really our job.

ensure_task.bat
PowerUser64, line 1: I don't think we should install things on the user's computer, that's not really our job.

README.md
```diff
@@ -102,3 +102,47 @@ To be able to build you will need a few things, depending on your OS
...
+
+```sh
+./ensure_task.sh [task arguments]
+```
+
+- If Go Task is not found, the script will attempt to install it using your system's package manager or by downloading a release.
+- Any arguments you provide will be passed directly to `task` (e.g., `run`, `clean`, etc.).
+- By default, running the script with no arguments will perform the default build.
+
+See the script files for more details or to customize the installation logic for your environment.
```
PowerUser64, line 148: (see previous comments about scripts)

Taskfile.yml
```diff
@@ -0,0 +1,193 @@
...
+    desc: Install tools on Windows (requires admin shell for winget/choco)
+    platforms: [windows]
+    cmds:
+      - cmd: winget install -e --id Kitware.CMake
+        ignore_error: true
+      - cmd: winget install -e --id Git.Git
+        ignore_error: true
+      - cmd: winget install -e --id Python.Python.3.12
+        ignore_error: true
+      - echo Ensure VS 2022 with "Desktop development with C++" workload is installed.
```
PowerUser64, line 46: Same as an earlier comment, I don't think we should install things on the user's computer.

PowerUser64 (commented):
I realize this is closed, but here are some comments. My main issue with this is I don't think we should install things on the user's computer, since I don't really think that's something for the build system to do.

Personally I prefer to use `just` over this (I authored #1922), but I would be open to alternatives if something else works better for some use cases.

davehorner:
thank you for the feedback.  I understand your concern with installing things.
I do like just.  if there were a just setup you can be sure I would be using it.
I enjoy using plain old Makefiles but lately I have been really enjoying Taskfile.yml.

The different points you make about usage or flags; this closed because it was not really looked at with an understanding of the system and usage.  I have recently started a working practice of trying to automate installs related to my live coding via things like Taskfile.yml and when doing so; its personal choice and I try my best to use conventions used...

>I don't think we should install things on the user's computer.

I think scripting out the entire process from checkout to run is a really valuable resource for users.  The project can decide to pick up as many different Makefile, Taskfile.yml. justfile. cargo make xtask whatever they like.  I feel comfortable taking on winget and choco deps and removing those invocations because they modify the system is not providing a working system from bare install. 

The Taskfile.yml - not a huge fan of yml, but a cross platform upgraded Makefile is welcome.

thanks again for the feedback and have a great night.

davehorner:
if it helps you can prompt. https://crates.io/crates/tidalcycles-rs

basically, if the taskfile doesn't go upstream, I still see myself doing something similar to the above crate for BespokeSynth.  It was just the other night I tried it and I'm SOOOO impressed.

Issue #1639, opened 2024-09-06
Songbuilder counts bars incorrectly when in "do not reset transport" mode.


n3wjack:
In the nightly build there is an option to switch off resetting the transport when the Songbuilder moves to a new scene.
This causes the Songbuilder module to misinterpret the number of bars it should play per scene.
To make it work, you have to add the number of bars you want to play, to the total amount of bars already played (known as the workaround for this issue).

## Build info

bespoke NIGHTLY 1.2.1 (Aug 19 2024 12:16:36) - Windows 10 - HEAD 8fa9efb

## Steps To Reproduce

- Add a Songbuilder to a song.
- Click the trangle button and set the option "reset_transport_every_sequencer_scene" to off.
- Create 2 scenes.
- Alternate between the two scenes, and set the bars to e.g. 2 or 4 for each step in the sequencer.
- Have the songbuilder play the song.

I attached a patch to demonstrate the issue.

[songbuilder-transport-bug.zip](https://github.com/user-attachments/files/16904612/songbuilder-transport-bug.zip)

![image](https://github.com/user-attachments/assets/6caca723-7a4f-4319-b1f3-c022ab654cbb)

## Expected vs. actual behavior

Expected:

For each step in the sequencer, the songbuilder plays the requested number of bars.

Actual:
After playing the first scene for the number of bars set, it plays only 1 bar for the subsequent scenes.

PR #1928, opened 2025-11-12
Fix bar counting in the SongBuilder when transport is off.

n3wjack:
Fix for issue #1639 

Added a new switch in the SongBuilder settings to be able to turn this fix on or off. 
This is on by default, so it won't break old projects. Turn it off to be able to enter bar values as when transport reset is off.

I don't know if this is a good way to do this, but is seemed like a logical place to put it, as it works together with the switch above it.

image

This project demonstrates the fix in the 2nd SongBuilder:

[songbuilder-transport-bug-v2.zip](https://github.com/user-attachments/files/23508612/songbuilder-transport-bug-v2.zip)

I'm not used to C++ development, so feel free to let me know if I did something incorrect here. 😉

### DIFF
--- a/Source/SongBuilder.cpp
+++ b/Source/SongBuilder.cpp
@@ -302,7 +302,10 @@ void SongBuilder::OnTimeEvent(double time)
 
    if (mSequenceStepIndex != -1)
    {
-      if (TheTransport->GetMeasure(time + TheTransport->GetListenerInfo(this)->mOffsetInfo.mOffset) >= mSequencerStepLength[mSequenceStepIndex] && !mSequencePaused)
+      int sequencerMeasure = GetSequencerMeasure();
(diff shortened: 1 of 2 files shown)


awwbees:
this doesn't appear to properly fix the issue, the progress bar on each row isn't filling as it should. I can take a look at a different fix. I also don't think it's important to provide an option to preserve the "legacy behavior", since the legacy behavior is just a straight-up bug

n3wjack:
I added the legacy behavior switch, so that old projects that have the workaround for ever-increasing bar values don't break if you it in a Bespoke version that has this fixed. But then again, you can always recalculate the values, and set the proper bar value.
The if-legacy check is kind of annoying though.

If you have a different fix, please do. Having to use a spreadsheet to calculate the bar values to enter is kind of annoying. 😁

Issue #474, opened 2021-10-24
[FR] `write audio` to support 16, 24, and float 32 wave audio format


AlexMitchellMus:
Support multiple audio wave output formats in BespokeSynth:

* 16 bit integer ±1.0
* 24 bit integer ±1.0
* 32 bit float unlimited

Specifically to support bypassing the limiter of float 32 bit audio saving.

Below is an idea for the layout of the signal path. In which the output audio of BespokeSynth would still be limited when sent to the output speakers.

By bypassing the limiter only for f32 wav, this will allow dynamics processing to be applied after the user has saved the audio file, hence allowing more freedom to experiment in bespoke, without having brick-wall limiting ruin improvised patching sessions.

For users that do not want to bypass the limiter- they would select a 16,24 bit integer audio format.

By definition, the float32 wav format would only be of benefit to the user if it is unlimited. A tooltip can explain the different use cases for different output audio formats.

![Screenshot from 2021-10-24 14-43-05](https://user-images.githubusercontent.com/12004932/138580446-ca63b7b7-fd04-4a9f-bfef-e2893ea2ba81.png)

AlexMitchellMus:
Having given this more thought, and discussion with another dev, I think BespokeSynth could safely do without the `limit` setting.

The system should be able to deal with larger than ±1 signals, and it's not desirable when using bespoke in a Jack solution, or in the future when Bespoke is a plugin.

jpcima:
As a note, the functionality can be implemented by a 3-choice setting of pcm16/pcm24/float32.
These would control the bit depth parameter of AudioFormatWriter.
If that bit depth is set to 32, juce selects a floating-point format.

https://github.com/BespokeSynth/BespokeSynth/blob/144de03b54f7a15aaa8ee54800810d77c4578b01/Source/Sample.cpp#L182

dmt-hub:
This would be great to have, is it still on the roadmap?

PR #1942, opened 2025-12-30
Support for 16, 24 & 32 bit WAV export.

n3wjack:
This satisfies feature request 474 except the limiter part, which isn't necessary IMO.
https://github.com/BespokeSynth/BespokeSynth/issues/474

Add a user preference settings to set the WAV export bit depth. 
Works for the global export audio button and the multitrack recorder. 
Changing the setting works for the next export without restarting Bespoke.

Screenshot of the setting dropdown:

image

### DIFF
--- a/Source/ModularSynth.cpp
+++ b/Source/ModularSynth.cpp
@@ -3533,7 +3533,7 @@ void ModularSynth::SaveOutput()
    auto outputTo = outputFile.createOutputStream();
    assert(outputTo != nullptr);
    bool b1{ false };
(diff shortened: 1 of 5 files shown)


PowerUser64:
Awesome! Tested it out and I was able to open 32bf files in audacity and hear values above 1.0f by using the amplify effect. I wonder if we could disable the output limiter (set it to zero) by default with this. Tbh I'm not really sure what purpose it serves since afaik the sound card should hard-clip everything no matter what. @awwbees do you have any input on this?

n3wjack:
Thanks for double-checking this! I checked the file types properties on Windows and that indicated that the files were indeed had different bit depths, and they all opened in Audacity, so I figured it was working.

PR #1945, opened 2026-01-09
Add `zerocrossrate`

merkys:
This PR introduces a `zerocrossrate` module which accepts incoming audio and outputs modulation values based on the measured [zero-cross rate](https://en.wikipedia.org/wiki/Zero-crossing_rate) using a sliding window approach. The module is really simple, but could be used to detect noise or other audio characteristics.


merkys:
Thanks for merging this @awwbees!

PR #1946, opened 2026-01-12
Fix FileChooser initial directory behavior for save/load state

westurner:
- Open to the directory of the current file and select the current file
- Refactor path logic to FileUtils:
  - Add FileUtils module (for getDirectoryOrDefault and other stateless testable file methods from ModularSynth.cpp)
  - Move path resolution logic to getDirectoryOrDefault in FileUtils
  - Update ModularSynth to use FileUtils::getDirectoryOrDefault
  - note about add'l stateless methods for FileUtils.cpp


westurner:
Perhaps for another PR:
- [ ] Make the default `savestate/` path a properly configurable parameter

westurner:
There's that additional global there for just the `"savestate/"` str, and also Ctrl-F `savestate/` shows a few paths in ModularSynth.cpp that would also need to be updated to be relative to a different path if specified

westurner:
I added a basic TestRunner in my build_bespoke repo (for building from source with a Fedora devcontainer) and wrote a test for this simple function: https://github.com/westurner/build_bespoke/blob/main/tests/SavestatePathTest.cpp

Which is probably overkill, but also how should unit and integration tests be added?

westurner:
Might as well add stdout logging of file opens and saves for debugging, too:

PR #1948, opened 2026-01-16
Add Harmonic Oscillator module

astro-mat:
This PR adds a new Harmonic Oscillator module to BespokeSynth.

The module has been:

- added to the project sources via CMake, and
- registered in ModuleFactory so it appears in the module menu.

I have built and tested this on Windows and confirmed that the module instantiates and produces audio.

This is my first contribution to BespokeSynth, so I’m very happy to make any requested changes or adjustments to fit project conventions.

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -237,6 +237,8 @@ target_sources(BespokeSynth PRIVATE
     GroupControl.h
     HelpDisplay.cpp
     HelpDisplay.h
+    HarmonicOscillator.cpp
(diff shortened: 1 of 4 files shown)


ArkyonVeil:
Greetings, can you post a video demo or say more about what it does? This PR says nothing about it besides the name.

awwbees:
this PR appears to be fully vibecoded, appears to not properly function (it's attempting to use an output cable as an input??), and I couldn't figure out how to use it, so I will decline to merge this

Issue #1300, opened 2023-06-23
Feat - parity between time interval selection menus


PowerUser64:
## Summary

There are many ways to select a time interval in bespoke, and not all of them have parity with each other, which can prove to be problematic when trying to automate changing time intervals across multiple modules.

Here are some modules and the time intervals they support.

| looper | pulser | notecounter | lfo |
| --- | --- | --- | --- |
| ![image](https://github.com/BespokeSynth/BespokeSynth/assets/24578572/ce359aba-f4e4-412e-bec0-2c936c544b05) |![image](https://github.com/BespokeSynth/BespokeSynth/assets/24578572/e24fbecc-d6e0-4bbb-a6c0-3f0fd90ccdf4) | ![image](https://github.com/BespokeSynth/BespokeSynth/assets/24578572/b199e511-d449-4427-9a41-2303b66f033c) | ![image](https://github.com/BespokeSynth/BespokeSynth/assets/24578572/b931dedb-c688-4fad-adbe-b829ccf3e92e) |

I understand the reasoning behind these menus - you're probably never going to want notes that are 64 bars in length from a notecounter, and it might be unusual to loop just one beat, but I think having a way to change these menus to specific values would be very valuable.

### Motivation

This would make it easier to make coordinated changes between multiple modules, without having to worry about figuring out which modules support which intervals.

KillyMXI:
> notes that are 64 bars

When note pitch is used to control radiosequencer for example.

NoxyNixie:
Sounds easier than one might expect. I've looked into this in the past and well, I discovered several bugs in the timing system within Bespoke that are still present to this day. To do this a lot of work and even more testing is required.

PR #1963, opened 2026-02-09
Implement free LFO interval mode in `tremolo`

merkys:
This PR implements "free" mode in `tremolo`, allowing to use different intervals than the built-in multiples and fractions of `1n`.

### DIFF
--- a/Source/TremoloEffect.cpp
+++ b/Source/TremoloEffect.cpp
@@ -42,11 +42,13 @@ void TremoloEffect::CreateUIControls()
    FLOATSLIDER(mAmountSlider, "amount", &mAmount, 0, 1);
    FLOATSLIDER(mOffsetSlider, "offset", &mOffset, 0, 1);
    FLOATSLIDER(mDutySlider, "duty", &mDuty, 0, 1);
+   FLOATSLIDER(mFreeRateSlider, "free rate", &mFreeRate, 0, 20);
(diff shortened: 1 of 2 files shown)


PowerUser64:
Great, this progresses toward #1300!

Tested, and it sounds nice! I think there's a slew limiter on the attack/release that keeps it from creating clicking noises. It's helpful on the lower speeds, but at higher speeds it makes for some interesting (possibly unintended) results. It seems like the internal oscillator has a lower sample rate than audio rate? If you turn on smoothing and sweep the rate upward, you'll hear some aliasing sounds as it goes higher (try putting the rate at 15999 hz while at 64/48000 buffer size+SR).

This is my guess as to the signal diagram of the module, based on what I'm hearing:
```
internal oscillator (aliasing happens here) -> slew limiter -> multiply with incoming signal
```

Tbh though, I think the slightly unusual characteristics are kind of nice. I like the softer sound you get from this at high speeds more than the sound you get out of LFO -> gain (at first I thought it would sound like LFO -> gain even at high speeds).

Also, minor thing, but I think it would probably be worth adding a curve to the rate slider so there's more room for adjustment closer to zero.

sound demo w/aliasing sounds: [tremolo-sine.mp3](https://github.com/user-attachments/files/25226517/tremolo-sine.mp3)

merkys:
Thanks for completing & merging this @awwbees!

PR #1996, opened 2026-03-28
website : double the « donated 20$ » checkmark for bespoke pro

AKArien:
If someone donates 40$, they have basically donated 20$ twice, so this deserves two checks in this box.

Just a little addition i felt would be fun to have :)


awwbees:
a nice idea, but I feel that this overcomplicates the joke :)

PR #2006, opened 2026-04-12
add pan for on_note script input

furrig:
This PR adds the missing pan parameter on script module input on_note.
It has been done so existing scripts will not break and has been tested with this in mind.

Documentation in /resource/scripting_reference.txt has been updated accordingly.

It was built and tested on Apple Silicon M1, macOS 26.4.1

First contribution here, looking forward to any corrections or pointers — happy to iterate.

### DIFF
--- a/Source/ScriptModule.cpp
+++ b/Source/ScriptModule.cpp
@@ -571,7 +571,7 @@ void ScriptModule::Poll()
          {
             //if (mPendingNoteInput[i].time < time)
             //   ofLog() << "trying to run script triggered by note too late!";
-            RunCode(mPendingNoteInput[i].time, "on_note(" + ofToString(mPendingNoteInput[i].pitch) + ", " + ofToString(mPendingNoteInput[i].velocity) + ")");
+            RunCode(mPendingNoteInput[i].time, "on_note(" + ofToString(mPendingNoteInput[i].pitch) + ", " + ofToString(mPendingNoteInput[i].velocity) + ", " + ofToString(mPendingNoteInput[i].pan) + ")");
          }
          mPendingNoteInput[i].time = -1;
       }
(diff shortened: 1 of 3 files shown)


awwbees:
can you describe a use case for this? it seems pretty obscure (I don't think I can think of a use case...) and it would be a shame to complicate the simple on_note() interface if there's no practical application

furrig:
My use case : 
- I use scripts to schedules notes
- I also have a script that maps notes to convert General Midi drums to DrumPlayer note numbers
- Then, in this scenario, when I use pan values on scheduling notes, I need it to go thru it all :)

Side note : 
- GM drums mapping allows me to play patterns using DrumPlayer or some other Drums machines without having to rewrite them 



As Pan is available with play_note, schedule_note, schedule_note_msg and note_msg methods, it seemed consistent to me that on_note would offer it too

awwbees:
thanks for the example. I'm going to have to think a little harder about if there's a cleaner way to pull this off... "pan" is only a sort-of-supported feature, and it feels like a shame to dirty up the clean `on_note(pitch, velocity)` API

awwbees:
for your specific use case, I wonder if `pitchremap` might be sufficient, without you having to write a custom script:
image

furrig:
Thanks for your answers

I did use pitchremaps at first but I needed several of them 'cos it's 8 notes max + that meant a lot of cables so finally a script was more convenient 

Waiting for your cleaner way ;)

awwbees:
just as a note, you can chain `pitchremap` like this, which might make it more appealing since it all stays linear:
image

furrig:
wow yes, thanks, haven't thought of that !
you saved me from a notesorter  :D

PR #2007, opened 2026-04-13
feat: do not include autosaves in the welcome screen

vrozkovec:
If autosaves are on, the welcome screen is cluttered with them and IMHO this defeats the purpose of recent files.
This change does not store autosaves in recent files, except if they were manually loaded by the user.

image

### DIFF
--- a/Source/ModularSynth.cpp
+++ b/Source/ModularSynth.cpp
@@ -3177,7 +3177,8 @@ void ModularSynth::LoadStatePopupImp()
 
 void ModularSynth::SaveState(std::string file, bool autosave)
 {
-   AddRecentFile(file, true);
+   if (!autosave)
+      AddRecentFile(file, true);
 
    mQueuedSaveStateInfo.mFile = file;
    mQueuedSaveStateInfo.mAutosave = autosave;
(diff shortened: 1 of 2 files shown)


PowerUser64:
It would be nice if there was a way to still show them, since it can make it faster to get back to where you were if you're recovering from a crash. Maybe there could be a checkbox called "show autosaves" near the top of the list that lets you enable showing them?

vrozkovec:
That would be fairly more complex and require another PR. But, I've changed the PR so it always shows the latest autosave as the first item, what do you think?

PowerUser64:
Hmm, one is much better than none (and should probably be enough in most crash-recovery cases), but I'd be concerned that by not showing all of them, people might not realize there are multiple autosaves.

This is probably good for now. I agree it's not great if we show all the autosaves at the top of the list, so I definitely want this to be considered either way. I'd be curious what @awwbees thinks.

A slight variation on what I suggested that might be smaller to implement could be to have a checkbox that changes between showing a list with _no_ autosaves (like you have it) and a list with _only_ autosaves. Just an idea though, do what you want w/it.

ArkyonVeil:
I'd lean that the latest autosave is a good start. Perhaps when hovering a file it shows a + button, which when clicked, expands the welcome to include all the auto saves of that file?

PR #2012, opened 2026-04-16
fix: envelope crash, curve memory corruption

vrozkovec:
On @ArkyonVeil 's impulse, I've fired up Claude Code to fix `envelope` crash when adding 10+ points and `curve` memory corruption that caused curve to be stuck with no way of bringing it back:
image

This is after the fix:
image

I just glanced over the changes and didn't have time to review them. Code compiles, and bugs are gone.
Review required.

Claude Code summary:
---

# Bug Fixes: Envelope Crash and Curve Corruption

## Bug 1: Envelope crash at 10+ points

The `EnvelopeEditor` has a UI controls array of size 10 (`mStageControls[10]`), but the backing ADSR data allows up to 20 stages (`MAX_ADSR_STAGES`). When you add an 11th point, `DrawModule()` at `EnvelopeEditor.cpp:511` accesses `mStageControls[10]` — out of bounds — and crashes.

## Bug 2: Curve module corruption at 20+ points

The "curve" module (`ModulatorCurve`) doesn't have `mStageControls`, so it survives past 10 points. But at 20 stages, the double-click "add point" code in `EnvelopeControl::OnClicked()` writes to `mStages[20]` (one past the 20-element array), causing memory corruption. This manifests as the garbled point positions you see in the screenshot — and can crash on some systems.

## Proposed fix (3 changes across 2 files)

1. **`EnvelopeEditor.h`** — increase `mStageControls` from 10 to `MAX_ADSR_STAGES` (20) so the UI matches the data capacity
2. **`EnvelopeEditor.cpp`** — add a bounds guard (if `numStages >= MAX_ADSR_STAGES`, skip insertion) before the add-point code
3. **`EnvelopeEditor.cpp`** — fix a delete-loop off-by-one that reads one-past-end, plus add a defensive clamp in `DrawModule()`

All four modules using `EnvelopeControl` (envelope, curve, velocitycurve, curvelooper) benefit from the fix.

---

# Fix: Envelope crash at 10+ points & Curve corruption at 20+ points

## Context

Two user-reported bugs in BespokeSynth's `EnvelopeControl` (shared by envelope, curve, velocitycurve, and curvelooper modules):

1. Envelope module crashes when adding 10+ control points (double-click to add)
2. Curve module corrupts visually / crashes for some users when adding 20+ points

Both stem from missing bounds checks in `EnvelopeControl::OnClicked()` and a size mismatch between `mStageControls[10]` and `MAX_ADSR_STAGES` (20).

### Crash at 10+ points (envelope editor only)

`EnvelopeEditor::DrawModule()` line 511 accesses `mStageControls[numStages - 1]` with no bounds check. `mStageControls` has 10 elements, so when `numStages > 10`, this is an out-of-bounds access.

### Corruption/crash at 20+ points (all EnvelopeControl users)

`EnvelopeControl::OnClicked()` line 213 — the "add stage on double-click" code shifts stages right starting from `i = GetNumStages()` and accesses `GetStageData(GetNumStages())`. When there are already 20 stages (`MAX_ADSR_STAGES`), this writes to `mStages[20]` — one past the end of the fixed-size array. Memory corruption follows.

### Delete loop off-by-one (latent bug)

Line 189 — the delete-stage shift loop runs `i < GetNumStages()`, reading `GetStageData(i + 1)` where `i+1 = GetNumStages()` on the last iteration. This reads one-past-end; if `GetNumStages() == MAX_ADSR_STAGES`, it's an out-of-bounds array access.

### 1. `EnvelopeEditor.h` line 139 — increase `mStageControls` to match `MAX_ADSR_STAGES`

```cpp
// Before:
std::array<StageControls, 10> mStageControls{};

// After:
std::array<StageControls, MAX_ADSR_STAGES> mStageControls{};
```

This ensures the UI can display controls for all possible stages. Adds ~400 bytes (negligible). Uses the named constant so the sizes stay in sync.

### 2. `EnvelopeEditor.cpp` — three changes

**A. Add bounds guard before inserting a new stage (lines 208–231)**

Wrap the existing insertion code in `if (mAdsr->GetNumStages() >= MAX_ADSR_STAGES) { reset highlight } else { existing code }`. This prevents the shift loop from ever writing past the array.

**B. Fix delete loop off-by-one (line 189)**

Change `i < mAdsr->GetNumStages()` to `i < mAdsr->GetNumStages() - 1`. The loop copies slot `i+1` into `i`; stopping one earlier avoids reading past the last valid stage.

**C. Defensive clamp in `DrawModule()` (line 511)**

Clamp the index used to access `mStageControls`:

```cpp
int lastVisibleStage = numStages - 1;
if (lastVisibleStage >= (int)mStageControls.size())
    lastVisibleStage = (int)mStageControls.size() - 1;
```

Even though change #1 makes this unreachable under normal operation, it prevents crashes from corrupted save files.

## Files to modify

- `/data/code/sources/BespokeSynth/Source/EnvelopeEditor.h` (line 139)
- `/data/code/sources/BespokeSynth/Source/EnvelopeEditor.cpp` (lines 189, 208, 511)

## Not changed

- `ADSRDisplay.h` `mDrawTimeHistory[10]` — this is a circular buffer for visual draw trail (10 frames), unrelated to stage count
- `ADSR.h` `GetStageData()` — no bounds check added; callers are now correct, and adding a clamp here would mask bugs

### DIFF
--- a/Source/EnvelopeEditor.cpp
+++ b/Source/EnvelopeEditor.cpp
@@ -186,7 +186,7 @@ void EnvelopeControl::OnClicked(float x, float y, bool right)
                mAdsr->GetStageData(mHighlightPoint + 1).time += mAdsr->GetStageData(mHighlightPoint).time;
                mAdsr->GetStageData(mHighlightPoint + 1).curve = mAdsr->GetStageData(mHighlightPoint).curve;
 
-               for (int i = mHighlightPoint; i < mAdsr->GetNumStages(); ++i)
+               for (int i = mHighlightPoint; i < mAdsr->GetNumStages() - 1; ++i)
                {
                   mAdsr->GetStageData(i).time = mAdsr->GetStageData(i + 1).time;
                   mAdsr->GetStageData(i).target = mAdsr->GetStageData(i + 1).target;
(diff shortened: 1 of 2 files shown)


ArkyonVeil:
Save/Load test is successful. Worth noting that this PR does not yet add unlimited/arbitrarily large cap of points to the affected modules. Simply fixes the crashing behavior, while adhering to the #define limit. 

Still the crash and memory corruption fixes are a very nice to have.

BespokeSynth_1fhMYac3Ze

PR #2013, opened 2026-04-16
feat: resizable curve

vrozkovec:
related to #2012 - without resizing, curve is unusable with more points on it


ArkyonVeil:
No issues saving/loading. Works as expected.
BespokeSynth_y2jvNZzt4Y

PR #2014, opened 2026-04-17
Organization: Moving the gigantic 10k line DrawFallbackText() function to its own file 

ArkyonVeil:
SynthGlobals.cpp is an important base file with many useful functions and tools. But there is a novella length, rarely used function in the middle of it, which makes the file annoying to parse.

This PR moves it to its own file. With no further changes besides a note in SynthGlobals.cpp noting the move.


awwbees:
resolved a slightly different way in #2022 (which avoided the situation where a function was declared in `x.h` and defined in `y.cpp`)

PR #2017, opened 2026-04-17
Feature: Sampler module now supports Pulse Input

ArkyonVeil:
Samplers now accept pulses and play the sample with a full velocity, pitch 48 note. (Affected by the Sampler's pitch value)

Supporting this was trivial and the pitch textbox was practically begging for pulser inputs.

...

Because one is quite overambitious, I decided to also throw in support for a couple pulse flags.

- Reset --> Stops all sound
- Repeat --> Interrupts the previous note and plays it on the same channel.
- Random --> Plays the sample with a random offset. Handy for granular play.

Repeat and Random are compatible and can optionally be layered.

https://github.com/user-attachments/assets/8f634f9a-474e-4cf7-bf01-7e28b84638d4

### DIFF
--- a/Source/Sampler.cpp
+++ b/Source/Sampler.cpp
@@ -271,7 +271,34 @@ void Sampler::OnPulse(double time, float velocity, int flags)
    //   note.velocity = velocity;
    note.velocity = 127;
    note.pitch = 48;
+
+   int repeatFlagInfo = -1;
+   if (flags & kPulseFlag_Reset) //Stops all sound.
+   {
+      mPolyMgr.KillAll();
+      note.velocity = 0;
+   }
+   else if (flags & kPulseFlag_Random) //Plays the sample at a random pos
+   {
+      repeatFlagInfo = mVoiceParams.mStartSample;
+      int range = mVoiceParams.mStopSample - mVoiceParams.mStartSample;
+      mVoiceParams.mStartSample = mVoiceParams.mStartSample + (gRandom() % range);
+   }
+   else if (flags & kPulseFlag_Repeat) //Replays on the last voice channel.
+   {
+      note.voiceIdx = mMostRecentVoiceIdx;
+   }
+   else if (flags & kPulseFlag_Backward)
+   {
+      //Obvious, but not supported. Pending polyphonic player/SampleVoice support.
+   }
+
    PlayNote(note);
+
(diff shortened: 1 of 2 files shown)


ArkyonVeil:
## IMPORTANT NOTE

After some more experimentation, with the Sampler + SongCanvasRack's implementation of it. It's clear that the pitch as displayed in this UI (and that gets passed to the internal polyphonic player) is not the same as the pitch values as passed by note. (It's likely reversed.) I'll update this when I manage to correct the SongCanvasRack's pitch play so it would match the notes being played.

ArkyonVeil:
## Update

Ran a couple tests. And... ultimately this is a bit of a weird situation. The PR is working exactly as designed, in that pulses are playing pitch 48 notes. On a base sample that is pitch shifted.

To actually make this work intuitively. In order that pulsing plays the sample with the note pitch seen in the sampler, WITHOUT breaking the current sample pitch internals... Then in an example samplePitch of 96, to play a Note Pitch of 96, and have it sound like they're 48 semitones apart like you'd expect. The note would have to be played at Pitch 144. Which raises faaar too many issues to explicitly support it like that.

I see a couple approaches:
- Leave it as is. It's technically working as designed, pitch shifting is just backwards for pulses if they're assumed to be note pitches. 
- Add a _second_ internal Polyphonic player designed just for pulses, which plays with the base sample pitch of 48. While this would allow pulsers to play samplers with the implied pitch in the Pitch textbox. Albeit since the textbox wouldn't be mapping directly to the polyphonic player's voice sample pitch. Live pitch shifting either wouldn't work, or require modifications to the polyphonic player to force it to work.
- Add a checkbox to shift the textbox maps to base sample pitch, or note pitch. (Seems extremely redundant since note inputs already do this with no textbox needed)
- Second textbox that maps this. Making the "simple" description in the tooltip even funnier.
- Eat the bullet and somehow make playing midi pitch notes higher than 127 possible, or at least invisibly supported in this one case.
- Rename the textbox/add a tooltip so the exact kind of pitch work its doing is clearer. 

In any regard, in my personal opinion it's still a net positive addition, even with this quirk. Perhaps a hypothetical future rework of this module could help clean details like these up. Don't really have much of an idea though.

Working on user feedback with the textbox's name and tooltip sounds like a good idea though. Ideas welcome.

PR #2019, opened 2026-04-18
Improvement: Welcome screen's demo files now have pictures.

ArkyonVeil:
From a new user perspective, only having the demo files for a description, with a placeholder logo on top; may discourage users from exploring the program.

Therefore I took several screenshots and framed them to be the most visually inviting in regard to the project's topic.

BespokeSynth_hShBpLY2W8


ArkyonVeil:
Also shortened the demo file names so they're a bit more readable. This only affects the welcome screen text name, the file names themselves are untouched.

BespokeSynth_NarTDDTR8E

Opinions welcome.

awwbees:
I like having screenshots for the examples, but I think this should be achieved just through re-saving the existing bsks (which will embed screenshot into the bsk), rather than special-casing it. similarly, I don't think there should be special-casing to show alias names for the files, and maybe instead the example files should just be renamed to fit better with the new welcome screen.

ArkyonVeil:
> I like having screenshots for the examples, but I think this should be achieved just through re-saving the existing bsks (which will embed screenshot into the bsk), rather than special-casing it. similarly, I don't think there should be special-casing to show alias names for the files, and maybe instead the example files should just be renamed to fit better with the new welcome screen.

Fair enough, albeit from a new user's perspective, would be having zoomed in screens be more effective to get them to try the example out? Rather than a zoomed out view of modules they do not know?

Just a thought.

Anyhow cheers.

Edit: Actually disregard this if it can be achieved by zooming.

PR #2033, opened 2026-04-28
Improvement: Dropdown Extensions

ArkyonVeil:
Currently, dropdown elements are used extensively throughout Bespoke. Despite this, they are seriously lacking in capabilities, Internally, a label only supports a name, and an index.

This is plenty for most option selections, but comes with several problems:

- Tooltips on labels aren't really supported, so the tooltip renderer is required to give special treatment to the dropdowns in the top bar. As well the EffectChain.
- If a developer wishes to provide non-module related tooltip information, this is currently not possible without significant workarounds. If they want their tooltips localized that's yet another problem.
- If a developer wishes to draw custom graphics on an item. This is not possible.
- Localized labels aren't a thing either, to the best of my knowledge. (Edit: Out of scope for this PR)

I intend to extend dropdowns so they can reasonably handle all of these cases while still largely using the same API.

I opted to create this PR early as a draft since it has far-reaching effects, and early notes would be greatly appreciated.

### Update 1

- Localization is seriously lacking in Bespoke. But I personally consider this out of scope for this PR. So my thoughts are to just extend the current tokenizer so dropdowns can support custom unique tooltips for now. This goes along the custom tooltip addresses, which will patch out the unique support required for dropdowns that spawn elements.

### Update 2

Should be ready for review.

Furthermore, I added two examples of the new features in use. As seen here:

https://github.com/user-attachments/assets/f0b2ac91-174b-4000-a60b-8efd9acc6664

Additionally, the effectchain, and the topbar now use the new API, with no need for hardcoded behavior in the ModularSynth.cpp file.

The new label struct:
```
struct DropdownListElement
{
   std::string mLabel; //Text on the dropdown
   int mValue{ 0 }; //Value of the dropdown, used for array indexing and modulation.

   //Custom tooltip address. Format: "module~control~subcontrol"
   //Can point to any tooltip in the tooltip files.
   //By default displays this if present --> thismodule~thisdropdown~thislabel~tooltip. If not present, no tooltip is shown.
   std::string mTooltipAddress{};

   float mReservedWidth{ 0.0f }; //Try to reserve additional space for special rendering
   DropdownRenderFn mRenderer{ nullptr }; //If set, will call this function on draw, for custom rendering. Still draws the label.
   std::string mRenderArgs; //Stores an argument for use. Sent into the function call.
};
```
Tokenizer also has been expanded to accept four tokens at once. Used for this. While I personally would like to do more, it's out of scope for this PR.

### Update 3

New format:

```
typedef void (*DropdownRenderFn)(ofRectangle renderRect, bool isHovering, bool isDrawingOnPopup, const DropdownListElement& element);
struct DropdownListElement
{
   std::string mLabel; //Text on the dropdown
   int mValue{ 0 }; //Value of the dropdown, used for array indexing and modulation.

   //Custom tooltip address. Format: "module~control~subcontrol"
   //Can point to any tooltip in the tooltip files.
   //By default displays this if present --> thismodule~thisdropdown~thislabel~tooltip. If not present, no tooltip is shown.
   std::string mTooltipAddress{};

   float mReservedWidth{ 0.0f }; //Try to reserve additional space for special rendering
   DropdownRenderFn mCustomRenderer{ nullptr }; //If set, will call this function on draw, for custom rendering. Still draws the label.
   void* mArgs {nullptr}; //Optionally include an argument for rendering work.
};
```

### DIFF
--- a/Source/DropdownList.cpp
+++ b/Source/DropdownList.cpp
@@ -221,7 +221,7 @@ void DropdownList::Render()
       ofPushMatrix();
       ofClipWindow(mX, mY, w - (mDrawTriangle ? 12 : 0), h, true);
       auto dropdownLabel = GetDisplayLabel(*mVar);
-      DrawTextNormal(GetDisplayValue(dropdownLabel.mValue), mX + 2 + xOffset, mY + 12);
+      DrawTextNormal(GetDisplayValue(*mVar), mX + 2 + xOffset, mY + 12);
       if (dropdownLabel.mRenderer)
       {
          dropdownLabel.mRenderer(ofRectangle(mX,mY,mWidth,mHeight),false,false,dropdownLabel.mRenderArgs);
--- a/Source/DropdownList.cpp
+++ b/Source/DropdownList.cpp
@@ -744,6 +744,14 @@ std::string DropdownListModal::GetHoveredLabel()
    return "";
 }
 
+DropdownListElement DropdownListModal::GetHoveredLabelObject()
+{
+   int index = mOwner->GetItemIndexAt((int)mMouseX, (int)mMouseY);
+   if (index >= 0 && index < mOwner->GetNumValues())
+      return mOwner->GetElement(index);
+   return {};
+}
+
 bool DropdownListModal::MouseMoved(float x, float y)
 {
    IDrawableModule::MouseMoved(x, y);
--- a/Source/DropdownList.cpp
+++ b/Source/DropdownList.cpp
@@ -220,7 +220,7 @@ void DropdownList::Render()
       DrawTextNormal(GetDisplayValue(*mVar), mX + 2 + xOffset, mY + 12);
       if (dropdownLabel.mCustomRenderer)
       {
-         dropdownLabel.mCustomRenderer(ofRectangle(mX, mY, mWidth, mHeight), false, false, dropdownLabel.mCustomRenderArgs);
+         dropdownLabel.mCustomRenderer(ofRectangle(mX, mY, mWidth, mHeight), false, false, dropdownLabel);
       }
       ofPopMatrix();
       if (mDrawTriangle)
@@ -340,7 +340,7 @@ void DropdownList::DrawDropdown(int w, int h, bool isScrolling)
 
       if (mElements[i].mCustomRenderer)
       {
-         mElements[i].mCustomRenderer(labelRect, isHovering, true, mElements[i].mCustomRenderArgs);
+         mElements[i].mCustomRenderer(labelRect, isHovering, true, mElements[i]);
       }
    }
    ofSetColor(255, 255, 255);
--- a/Source/DropdownList.cpp
+++ b/Source/DropdownList.cpp
@@ -73,7 +73,6 @@ void DropdownList::AddLabel(const DropdownListElement& item)
    CalculateWidth();
    mHeight = kItemSpacing;
    CalcSliderVal();
-   mIndexDisplayCache = -1;
 }
 
 void DropdownList::RemoveLabel(int value)
@@ -86,7 +85,6 @@ void DropdownList::RemoveLabel(int value)
 
          CalculateWidth();
          mHeight = kItemSpacing;
-         mIndexDisplayCache = -1;
          CalcSliderVal();
          break;
       }
@@ -103,7 +101,6 @@ void DropdownList::SetLabel(const std::string& label, int value)
 
          CalculateWidth();
          mHeight = kItemSpacing;
-         mIndexDisplayCache = -1;
          CalcSliderVal();
          return;
       }
@@ -120,7 +117,6 @@ void DropdownList::SetLabel(const DropdownListElement& item, int value)
 
          CalculateWidth();
          mHeight = kItemSpacing;
-         mIndexDisplayCache = -1;
          CalcSliderVal();
          return;
       }
@@ -222,9 +218,9 @@ void DropdownList::Render()
       ofClipWindow(mX, mY, w - (mDrawTriangle ? 12 : 0), h, true);
       auto dropdownLabel = GetDisplayLabel(*mVar);
       DrawTextNormal(GetDisplayValue(*mVar), mX + 2 + xOffset, mY + 12);
-      if (dropdownLabel.mRenderer)
+      if (dropdownLabel.mCustomRenderer)
       {
-         dropdownLabel.mRenderer(ofRectangle(mX, mY, mWidth, mHeight), false, true, dropdownLabel.mRenderArgs);
+         dropdownLabel.mCustomRenderer(ofRectangle(mX, mY, mWidth, mHeight), false, false, dropdownLabel.mCustomRenderArgs);
       }
       ofPopMatrix();
       if (mDrawTriangle)
@@ -342,9 +338,9 @@ void DropdownList::DrawDropdown(int w, int h, bool isScrolling)
 
       DrawTextNormal(mElements[i].mLabel, 1 + mMaxItemWidth * col, (i % maxPerColumn) * kItemSpacing + 12 + pageHeaderShift);
 
-      if (mElements[i].mRenderer)
+      if (mElements[i].mCustomRenderer)
       {
-         mElements[i].mRenderer(labelRect, isHovering, false, mElements[i].mRenderArgs);
+         mElements[i].mCustomRenderer(labelRect, isHovering, true, mElements[i].mCustomRenderArgs);
       }
    }
    ofSetColor(255, 255, 255);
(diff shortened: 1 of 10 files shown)


Source/DropdownList.h
```diff
@@ -197,6 +211,7 @@ class DropdownList : public IUIControl, public IPulseReceiver
    bool mAutoCalculateWidth{ false };
    bool mDrawTriangle{ true };
    double mLastScrolledTime{ -9999 };
+   mutable int mIndexDisplayCache{ -1 };
```
awwbees: did you make this optimization based upon some sort of profiling? I wouldn't think so... FindItemIndex() is very simple and fast. if you didn't observe any specific performance issues that you're fixing, I'd say: this caching optimization introduces unnecessary maintenance complexity for negligible performance gains, and should be removed

Source/SongBuilder.cpp
```diff
@@ -1122,10 +1122,51 @@ void SongBuilder::ControlTarget::CreateUIControls(SongBuilder* owner)
...
 }
 
+void SongBuilder::DrawColourCircle(ofRectangle renderRect, bool isHovering, bool drawnOnBody, const std::string& args)
+{
+   if (drawnOnBody)
+      return;
+   ofPushStyle();
+
+   ofColor drawCol;
+   if (args == "grey")
```
awwbees: there's already a mapping of these strings to colors in mColors, this lookup should happen with that rather than creating a parallel mapping

Source/DropdownList.h
```diff
@@ -29,12 +29,21 @@
 #include "IDrawableModule.h"
 #include "ClickButton.h"
 #include "IPulseReceiver.h"
-#include "PatchCableSource.h"
 
+typedef void (*DropdownRenderFn)(ofRectangle renderRect, bool isHovering, bool drawnOnButtonBody, const std::string& args);
```
awwbees: "drawnOnButtonBody" is unclear naming, and seems to be inconsistently named elsewhere ("drawnOnBody" and "isDrawnOnDropdownList"), this should have a clearer name for its purpose and be made consistent across this PR

Source/DropdownList.cpp
awwbees: this should be refactored to call the new GetHoveredLabelObject() below, and return the mLabel from what it returns

resource/tooltips_eng.txt
```diff
@@ -1991,6 +1991,43 @@ splitter~splits a stereo signal into two mono signals, or duplicates a mono sign
 scale~controls the global scale used by various modules
 ~root~root note of the scale
 ~scale~which set of notes to use
+~~ionian~1 2 3 4 5 6 7\n\nmood: bright, balanced, clear, confident, resolved\ngenres: pop, classical, folk, singer-songwriter, country
```
awwbees: double-tilde seems like odd/broken notation, shouldn't it be something like `~scale~ionian~`? maybe a wildcarded `~*~ionian~` in the case that there are multiple controls within a module that share tooltips?

resource/tooltips_eng.txt
```diff
@@ -1991,6 +1991,43 @@ splitter~splits a stereo signal into two mono signals, or duplicates a mono sign
...
+~~aeolian~1 2 b3 4 5 b6 b7\n\nmood: dark, melancholic, natural, introspective, emotional\ngenres: rock, pop, metal, folk, indie
```
awwbees: I don't feel comfortable with bespoke ascribing values or genres to scale modes, these should be removed

Source/DropdownList.h
```diff
@@ -29,12 +29,21 @@
...
+   std::string mLabel; //Text on the dropdown
+   int mValue{ 0 }; //Value of the dropdown, used for array indexing and modulation.
+
+   //Custom tooltip address. Format: "module~control~subcontrol"
+   //Can point to any tooltip in the tooltip files.
+   //By default displays this if present --> thismodule~thisdropdown~thislabel~tooltip. If not present, no tooltip is shown.
+   std::string mTooltipAddress{};
+
+   float mReservedWidth{ 0.0f }; //Try to reserve additional space for special rendering
+   DropdownRenderFn mRenderer{ nullptr }; //If set, will call this function on draw, for custom rendering. Still draws the label.
```
awwbees: maybe call this mIconRenderer or mCustomRenderer

resource/tooltips_eng.txt
ArkyonVeil: Hm? Thought mood at least would be a bit handy for newcomers since they don't know how to interpret scales and how they're typically used.

Either way, this was just an example. Can always be revisited later.

So just the keys alone?

resource/tooltips_eng.txt
ArkyonVeil: I designed the format with nested logic.
0 = module
1 = ~controlofmoduleabove
2 = ~~subcontrolofcontrolofmoduleabove
A rough extension of the current system. 

`~scale~ionian~` is technically also valid. Albeit since 'scale' is ignored by the tokenizer in this example. It could cause tooltip writers to misfile. Hence why the example omits it, and "`~~`" refers to the control above it.

If a developer currently wants to reuse tooltips, they certainly still can with tooltipAdresses since they work in a nested context. (Albeit these would need to be manually corrected if the element is renamed and they hardcode the reference rather than pulling it from constructed data)

`~*~thing~ `Is not a bad idea though. Albeit requires reworking the tokenizer a bit.

Source/DropdownList.h
ArkyonVeil: I renamed this twice and I still dunno what to call it. I'll take another look.

ArkyonVeil:
Went through all changes (plus added an example of generic subcontrol tooltips on the eq module). I'd like to note two things though:

- The Dropdown argument I chose, std::string was a bit inspired by some of my former work abusing C# strings for fun. But since they store bytes in C++ land. This makes them extremely limited as a choice for arguments container. Albeit not quite sure what I should replace them with. What's the usual for essentially an arbitrary container?
- I stripped the info on scales as requested. But without any info on scales. For a user who doesn't know what the scale already does, only having the notes as information is largely useless. _(Ahem! Me, I'm one of those users.)_

resource/tooltips_eng.txt
awwbees: maybe just the keys alone, and add "(major)" and "(minor)" to the modes that it applies to ("major" for ionian, lydian, and mixolydian, "minor" for aeolian, dorian, phrygian, and locrian)

mood and genre, especially in the context of just a set of notes, is a very subjective thing, and I don't think it's responsible for bespoke to present that as any sort of "truth". besides, over half of the scales are described as "jazz" or "jazzy", which makes it kinda useless and speaks to the specious nature of genre-labeling

ArkyonVeil: Done

resource/tooltips_eng.txt
ArkyonVeil: Already done, pending review.

awwbees:
> * The Dropdown argument I chose, std::string was a bit inspired by some of my former work abusing C# strings for fun. But since they store bytes in C++ land. This makes them extremely limited as a choice for arguments container. Albeit not quite sure what I should replace them with. What's the usual for essentially an arbitrary container?

it seems like an arbitrary container isn't needed here, and instead the renderer just needs to know which element it's drawing for. rather than creating this args param, maybe it'd be better if it just passes along which `DropdownListElement` it's drawing: `typedef void (*DropdownRenderFn)(ofRectangle renderRect, bool isHovering, bool isDrawingOnPopup, const DropdownListElement& element);`

the owner/renderer has all the context it needs to figure out what to draw, so there's no need to pass along redundant parameters. for example, `SongBuilder::DrawColourCircle()` can look up which color to draw based upon the element's label

ArkyonVeil:
Updated the format, best of both worlds, now includes its own label. Which itself includes an optional argument which can be a pointer to an array, or the module itself, or whatever is necessary.

```
typedef void (*DropdownRenderFn)(ofRectangle renderRect, bool isHovering, bool isDrawingOnPopup, const DropdownListElement& element);
struct DropdownListElement
{
   std::string mLabel; //Text on the dropdown
   int mValue{ 0 }; //Value of the dropdown, used for array indexing and modulation.

   //Custom tooltip address. Format: "module~control~subcontrol"
   //Can point to any tooltip in the tooltip files.
   //By default displays this if present --> thismodule~thisdropdown~thislabel~tooltip. If not present, no tooltip is shown.
   std::string mTooltipAddress{};

   float mReservedWidth{ 0.0f }; //Try to reserve additional space for special rendering
   DropdownRenderFn mCustomRenderer{ nullptr }; //If set, will call this function on draw, for custom rendering. Still draws the label.
   void* mArgs {nullptr}; //Optionally include an argument for rendering work.
};
```

PR #2034, opened 2026-05-03
Improvement: Unlimited Envelopes + Envelope Editor Rework  - Design Proposal, Discussion Welcome!

ArkyonVeil:
For several years now, envelopes have effectively been a handful of points away from crashing Bespoke. A hard cap is set at 20, but oversights mean that it can crash at around 10 points. This is an extremely small limit for an envelope system. When it doesn't crash, it malfunctions and corrupts the current envelope.

My personal motivations are as an enthusiast in heavy EDM music. Half-broken envelopes are a key issue to resolve. As the envelopes common in that style can be highly detailed and of great length. Things not quite possible in current versions.

Therefore the goal of this PR is to resolve these issues, while reworking the envelope editor to cleanly support envelopes of arbitrary size.

### Obstacles
- Performance, as the audio thread is of key importance. Minimizing costs in every occasion is paramount, as lag is directly reflected on the user's work.
- The limits are hardcoded, and several parts of both the ADSR system and the envelope editor work on this assumption, IE, pre-making arrays of UI elements for all 20. This needs to be reworked.
- The Envelope editor is largely built on the assumption of only having 20 points maximum. Any point added, it draws additional point editors. While this is effective in small limits. It gets increasingly unworkable for dense workspaces, even with a scrollbar. 

### Goals
- Fixing crashes (largely done by @vrozkovec 's #2012 )
- Outright removal of the ADSR/Envelope limit
- Minimizing performance costs of this improvement by making use of hybrid stack/heap iteration techniques.
- Overhauling the envelope editor so it cleanly supports arbitrarily dense envelopes.
- Maintaining save compatibility.

### Design
Designer_jCLg6VQbeB

- Now by default, editable/patchable options are only available if the user double-clicks or shift-clicks a point on the envelope. Expanding the workspace vertically and drawing editable controls.
- "Open-menu" points have their IDs highlighted in yellow for contrast.
- These controls can be deleted by either removing the point itself, or click the (X) on the menu.
- Their IDs are ordered by horizontal left-to-right position, incremented for every  "open-menu" point. This id will shift if a new point is place in between or before.
- Point feedback is drawn on the bottom left. (Or top left if the user is hovering that area), they are only drawn if the user is hovering over a line or a point.
- Scrollbars are drawn if the amount of information necessary exceeds the workspace. This avoid having to resize the workspace to edit previously offscreen options.
- Additional highlights for user feedback will be drawn when applicable.

As this is a WIP project with far-reaching effects. Therefore I have opted to once again, perform an early draft pull request. Early feedback is welcome as it helps to narrow and focus development to have a more useful envelope system for everyone.

### DIFF
--- a/Source/ADSR.cpp
+++ b/Source/ADSR.cpp
@@ -29,7 +29,7 @@
 #include "FileStream.h"
 #include "SynthGlobals.h"
 
-void ::ADSR::Set(float a, float d, float s, float r, float h /*=-1*/)
+void ADSR::Set(float a, float d, float s, float r, float h /*=-1*/)
 {
    mStages[0].target = 1;
    mStages[0].time = MAX(a, 1);
@@ -46,7 +46,7 @@ void ::ADSR::Set(float a, float d, float s, float r, float h /*=-1*/)
    mHasSustainStage = true;
 }
 
-void ::ADSR::Set(const ADSR& other)
+void ADSR::Set(const ADSR& other)
 {
    for (int i = 0; i < other.mNumStages; ++i)
       mStages[i] = other.mStages[i];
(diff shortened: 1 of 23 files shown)


ArkyonVeil:
Just to disclose some additional bias. Fixed envelopes are an absolute requirement for the SongCanvas project. The workflow is almost guaranteed to cause crashes without something like this.

PR #2035, opened 2026-05-03
Feature: Unsafe Cable Connections

ArkyonVeil:
Currently, there is an oversight in cable patching the float slider. In that elements that are marked as noHover, and can't be currently connected, still allow this.

While I personally prodded on interest to have this oversight patched, an user indicated that in certain cases, users may desire the upmost freedom to undergo any risk for whatever they're messing with at the time. Even if the use of that option is not typically intended by the developer.

So, in tandem, I thought about officially supporting this. By the means of adding a dedicated unsafe cable connections mode, which allows even expressly disallowed connections.

BespokeSynth_AOQOBxpPVe

This PR adds a couple things:

- Adds an option for unsafe_cable_connections
- This allows for connecting cables even when 'unsafe' as well as formally allowing unsafe LFO attachments.
- Unsafe controls are drawn in orange, and attached LFO attempts flash briefly before allowing with another click.
- These interactions are only visible when unsafe is on. If off, the default, is to not render any hover highlights as previous.
- Affects disallowed connections via GetCableTargetable() and GetNotHover
- Unpatchable IUIcontrols hover in white.

However, since this allows the user to do things which may cause instability. I opted to add the following for more extreme cases:
- An extension to IUIControl, which allows the developer to flag controls using SetCableTargetableBlockHard().
- This overrides all unsafe cable patching options. And is meant for cases where misuse may cause instability, crashing or corruption.
- The ModuleSaveDataPanel has this enabled by default.

### DIFF
--- a/Source/IUIControl.cpp
+++ b/Source/IUIControl.cpp
@@ -29,6 +29,7 @@
 #include "ModularSynth.h"
 #include "PatchCable.h"
 #include "Push2Control.h"
+#include "UserPrefs.h"
 
 //static
 IUIControl* IUIControl::sLastHoveredUIControl = nullptr;
@@ -117,9 +118,28 @@ void IUIControl::DrawHover(float x, float y, float w, float h)
    }
    else if (gHoveredUIControl == this && IKeyboardFocusListener::GetActiveKeyboardFocus() == nullptr && TheSynth->GetGroupSelectedModules().empty())
    {
+      double time = gTime / 1000.0f;
       ofPushStyle();
       ofNoFill();
-      ofSetColor(0, 255, 255, 255);
+      if (mHoverWarningFlashTime < time)
+      {
+         if (mCableTargetable == 1)
+            ofSetColor(0, 255, 255, 255);
+         else
+         {
+            ofSetColor(255, 255, 255, 255);
+         }
+      }
+      else
+      {
+         int iTime = static_cast<int>(time * 4);
+
+         iTime = iTime % 2;
+         if (iTime == 0)
+            ofSetColor(255, 255, 255, 0);
+         else
+            ofSetColor(mHoverWarningColour);
+      }
       ofRect(x, y, w, h, 4);
       ofPopStyle();
    }
@@ -161,7 +181,10 @@ void IUIControl::DrawPatchCableHover()
       ofPushStyle();
       ofNoFill();
       ofSetLineWidth(1.5f);
-      ofSetColor(255, 0, 255, 200);
+      if (mCableTargetable == 1)
+         ofSetColor(255, 0, 255, 200);
+      else
+         ofSetColor(255, 115, 0, 235);
       ofRect(mX, mY, w, h);
       ofPopStyle();
    }
@@ -169,9 +192,7 @@ void IUIControl::DrawPatchCableHover()
 
 bool IUIControl::CanBeTargetedBy(PatchCableSource* source) const
 {
-   if (!mCableTargetable)
-      return false;
-   if (GetNoHover())
+   if (!GetCableTargetable())
       return false;
    return source->GetConnectionType() == kConnectionType_Modulator || source->GetConnectionType() == kConnectionType_ValueSetter || source->GetConnectionType() == kConnectionType_UIControl;
 }
@@ -184,6 +205,20 @@ void IUIControl::StartBeacon()
       moduleParent->StartBeacon();
 }
 
+bool IUIControl::GetCableTargetable() const
+{
+   if (!UserPrefs.unsafe_cable_connections.Get())
+   {
+      if (!mCableTargetable || GetNoHover())
+         return false;
+   }
+   else
+   {
+      if (mCableTargetable == -1)
+         return false;
+   }
+   return true;
+}
 void IUIControl::PositionTo(IUIControl* anchor, AnchorDirection direction)
 {
    ofRectangle rect = anchor->GetRect(true);
@@ -407,4 +442,9 @@ void IUIControl::DestroyCablesTargetingControls(std::vector<IUIControl*> control
    }
    for (const auto cable : cablesToDestroy)
       cable->Destroy(false);
+}
+void IUIControl::TriggerHoverWarning(ofColor colour, float seconds)
+{
+   mHoverWarningColour = colour;
+   mHoverWarningFlashTime = gTime / 1000.0 + static_cast<double>(seconds);
 }
--- a/Source/IUIControl.h
+++ b/Source/IUIControl.h
@@ -87,7 +87,7 @@ public:
    //Enable/disable cable patching on compatible cables
    //Can be user overridden by ticking unsafe_cable_patching
    void SetCableTargetable(bool targetable) { mCableTargetable = static_cast<int>(targetable); }
-   void SetCableTargetableBlockedHard() { mCableTargetable = -1; } //Disallows cable patching, even unsafe.
+   void SetCableTargetableBlockHard() { mCableTargetable = -1; } //Disallows cable patching, even unsafe.
    bool GetCableTargetable() const;
 
    void SetNoHover(bool noHover) { mNoHover = noHover; }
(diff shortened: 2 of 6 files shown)


ArkyonVeil:
I'm thinking that red (instead of orange) might be more indicated for a 'dangerous' operations, I can't seem to make up my mind. Feedback welcome.

PowerUser64:
I don't think there really _is_ a color you could pick that users would immediately understand what it's indicating. Since, in my mind at least, that part of "the language of the UI" doesn't really have the ability to say what you're trying to put there. I think you need to use something different than an outline to express this.

If you put a warning icon (⚠️) there, it would probably suffice. The warning icon is currently used when you attach a stereo cable to a mono input, and I think it would make sense to use it here as well. Having a tooltip to describe what "unsafe" means in this context would probably be helpful too.

ArkyonVeil:
> I don't think there really _is_ a color you could pick that users would immediately understand what it's indicating. Since, in my mind at least, that part of "the language of the UI" doesn't really have the ability to say what you're trying to put there. I think you need to use something different than an outline to express this.
> 
> If you put a warning icon (⚠️) there, it would probably suffice. The warning icon is currently used when you attach a stereo cable to a mono input, and I think it would make sense to use it here as well. Having a tooltip to describe what "unsafe" means in this context would probably be helpful too.

Having these options available does require the unsafe option toggled on. Unfortunately, since dynamic tooltips aren't really a thing (plus they can be disabled), it's up to developers to partially cover this by writing tooltips their correctly. Albeit, with a warning sign, that could probably work.

PowerUser64:
Oh! I misunderstood your initial comment. If unsafe is a setting, then you are right: an outline does seem to make sense. I thought you meant unsafe would be an internal thing that could be set on sliders (that users can't control). My bad!

In that case, I think orange is a great color. In fact, I think it would be best not to use red, since red is used for 'error cables' (nan/±inf values) and that expression of "this thing is ACTUALLY broken and you need to do something" might be useful for sides at some point too. That's just my take though, there could be something else I'm not thinking of. What do you think?

ArkyonVeil:
> Oh! I misunderstood your initial comment. If unsafe is a setting, then you are right: an outline does seem to make sense. I thought you meant unsafe would be an internal thing that could be set on sliders (that users can't control). My bad!
> 
> In that case, I think orange is a great color. In fact, I think it would be best not to use red, since red is used for 'error cables' (nan/±inf values) and that expression of "this thing is ACTUALLY broken and you need to do something" might be useful for sides at some point too. That's just my take though, there could be something else I'm not thinking of. What do you think?

I do lean a bit on "what if the user forgets", but then again. These outlines only appear in cable dragging contexts, and explicitly require unsafe_cable_connections to be turned on.

Also, while this might be a feature that sounds dangerous. It's still up to developers to allow this, since cable blocking can still be overridden by using SetCableTargetableBlockHard(). It's more of a soft discouragement.

Albeit as is, there are probably a lot of cases where this is indeed dangerous, so might require further patching down the line. Either way, it IS unsafe. <>P


The point of this is to fix an oversight. And say to the user "You probably don't need to patch this and will likely mess your module. But if you really want to..."

afentoe:
Sounds like a great feature.

My suggestion:
Always allow dangerous connections, but activate dangerous connections by holding an activation key when starting a new connection. (Control-key perhaps? That would act similar to control-click in a slider, to modify the limits: potentially also very dangerous)
Shift+Control combi would add a new dangerous connection to an existing cable point.

When dangerous cable is active: Show different colors for destinations: orange for dangerous connection sounds good.
Optional (perhaps not needed as user is a poweruser and orange color indication is enough):
When hovering over an orange (dangerous) destination, show a tooltip with a warning, or show the warning icon, or perhaps show the warning in the top left, (same place where the path is shown when hovering)

What do you think about the control-key option to activate dangerous connections when clicking a cable point?


Regarding a setting to activate dangerous connections:
Using a global option to enable or disable this functionality, might get a bit messy.

For instance:
How should unsafe links behave, when another user, who has unsafe connections set to off, loads a patch with unsafe connections.
I think unsafe cables should always remain connected when opening, but next issue: how to handle editing the unsafe connection in that case:
- Fixed/non-editable?
- Remove only? (how would that work if dragging a non-safe cable?)

Do you intend for the Unsafe cable option, to be a global option, a patch related option (how?), or perhaps a module specific setting (triangle menu)?

So perhaps the solution to not use an option toggle, but only a modifier key, is way easier to handle all these situations.

ArkyonVeil:
> Sounds like a great feature.
> 
> My suggestion: Always allow dangerous connections, but activate dangerous connections by holding an activation key when starting a new connection. (Control-key perhaps? That would act similar to control-click in a slider, to modify the limits: potentially also very dangerous) Shift+Control combi would add a new dangerous connection to an existing cable point.
> 
> When dangerous cable is active: Show different colors for destinations: orange for dangerous connection sounds good. Optional (perhaps not needed as user is a poweruser and orange color indication is enough): When hovering over an orange (dangerous) destination, show a tooltip with a warning, or show the warning icon, or perhaps show the warning in the top left, (same place where the path is shown when hovering)
> 
> What do you think about the control-key option to activate dangerous connections when clicking a cable point?
> 
> Regarding a setting to activate dangerous connections: Using a global option to enable or disable this functionality, might get a bit messy.
> 
> For instance: How should unsafe links behave, when another user, who has unsafe connections set to off, loads a patch with unsafe connections. I think unsafe cables should always remain connected when opening, but next issue: how to handle editing the unsafe connection in that case:
> 
>     * Fixed/non-editable?
> 
>     * Remove only? (how would that work if dragging a non-safe cable?)
> 
> 
> Do you intend for the Unsafe cable option, to be a global option, a patch related option (how?), or perhaps a module specific setting (triangle menu)?
> 
> So perhaps the solution to not use an option toggle, but only a modifier key, is way easier to handle all these situations.

Hmm, it's very much intentionally a global override, as it allows for patching which the developer deems unnecessary and might have not stress tested it with the input dynamics of an LFO or modulator. It might also just require some hacky behavior to work, like the notecanvas. (It's internally wired to disable LFOs, so it requires a middleman modulator to work), and saving might not work perfectly either.

Basically, it's for people who want to break things, and it takes a couple guard rails off. Like I mentioned before. (Before you could just right click to bypass this, which seems... problematic)

> Always allow dangerous connections, but activate dangerous connections by holding an activation key when starting a new connection. (Control-key perhaps? That would act similar to control-click in a slider, to modify the limits: potentially also very dangerous)

From a keybind perspective... That's actually dangerous, Bespoke is already an unusual program, and as people learn and press random keys. They might accidentally enable this, not know what it means, and risk losing a ton of work. The flag's name includes unsafe and is buried in options, so people are unlikely to turn it on carelessly.

> When hovering over an orange (dangerous) destination, show a tooltip with a warning, or show the warning icon, or perhaps show the warning in the top left, (same place where the path is shown when hovering)

The yellow warning is fine. Perhaps unnecessary, but I wouldn't disagree with its inclusion.

> How should unsafe links behave, when another user, who has unsafe connections set to off, loads a patch with unsafe connections.

IIRC given the way it works, the flag bypasses conditions that usually fail (disallow hover/nopatching) when you drag things. So this likely wouldn't affect patches at all (unsafe cables remain attached, baring unsupported weirdness). Albeit, I haven't tested it.

In my opinion it's fine. Since you can't rewire them anyway if the option is off.

> I think unsafe cables should always remain connected when opening, but next issue: how to handle editing the unsafe connection in that case:

What I'd expect is that you can rewire, but can't reattach, but if you cancel it resets fine. I haven't tested this since I don't use patches.

>Do you intend for the Unsafe cable option, to be a global option, a patch related option (how?), or perhaps a module specific setting (triangle menu)?

Global, per module setting it seems a bit too generic and will add noise to every module option in Bespoke's roster. Plus, it's more of a user preference really. IE, Are you happy with 95% freedom, or do you want 99% freedom? (The last 1% is the hard patch blockers on the options menu so you don't accidentally corrupt your module/project.)

#### Long Story Short

This PR patches the right-click to spawn an LFO on blocked controls and adds an option to turn it back on. Plus gives you freedom to employ the same breakage with direct cable patching as well on any other module, responsibility is on the user. That's pretty much it. <>)

PR #2036, opened 2026-05-04
Optimization: Sea of Grain

ArkyonVeil:
For a while now, every time I added a Sea of Grain module, my framerate has consistently dropped by a third. Which is impressive for a module that draws so little when brand new.

After some profiling, I came into the conclusion that... to use an industry term, "pisses away" frames doing pretty much nothing at all.

So I ended up doing some aggressive culling to cut down on the amount of drawing as much as possible + batching some operations when they do not interfere. The end result is about a 90%-95%~ draw framerate increase. Even when all controls are active and blasting.

### Before

BespokeSynth_fEr4owhRe6

### After

BespokeSynth_RXVq7sivnj

An unrelated change is that by default, only the first set of controls is shown to the user. This improves performance to a minor extent and makes the module look significantly less intimidating.

### DIFF
--- a/Source/SeaOfGrain.cpp
+++ b/Source/SeaOfGrain.cpp
@@ -50,7 +50,11 @@ SeaOfGrain::SeaOfGrain()
       mMPEVoices[i].mOwner = this;
 
    for (int i = 0; i < kNumManualVoices; ++i)
+   {
       mManualVoices[i].mOwner = this;
+      if (i >= 1)
+         mManualVoices[i].mGranulator.mSpawnGrains = false;
+   }
 }
 
 void SeaOfGrain::CreateUIControls()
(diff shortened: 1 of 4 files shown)


aulerius:
doing god's work 💪

Source/SeaOfGrain.cpp
```diff
@@ -369,6 +385,7 @@ void SeaOfGrain::LoadFile()
 
 void SeaOfGrain::ButtonClicked(ClickButton* button, double time)
 {
+
```
PowerUser64: Did you mean to add this newline here? There isn't really anything wrong with  it, but it caught my eye (this one and the one at the start of `GrainManualVoice::Draw` especially since there are no other changes in these fn's).

PowerUser64 (approved):
So the issue was that all voices were always drawing their grains no matter what? XD

Source/SeaOfGrain.cpp
```diff
@@ -503,6 +545,11 @@ void SeaOfGrain::LoadState(FileStreamIn& in, int rev)
       mSample->LoadState(in);
       UpdateSample();
    }
+
+   for (int i = 0; i < kNumManualVoices; ++i)
+   {
```
merkys, line 546: Curly brackets are not needed for this `for`

Source/SeaOfGrain.cpp
ArkyonVeil: I wrote a lot of profiling code, but probably left a bit of dust on the way.

PR #2050, opened 2026-05-17
allow minimum pitchbend range to be 0 in midicontroller

### LINKED ISSUE #2015: Add ability to ignore pitch bend or set its range to zero
## Summary

This feature could be achieved trough one of the following:

1. Allow for `pitchbendrange` on `midicontroller` to be set at 0, which would effectively disable it.
2. Allow `pitchbender` module to also **set** or **multiply** the incoming pitch bend information so that it would be possible to be nullified to zero.

### Motivation

Midi keyboards that have faulty, broken, or glitchy pitch bend joysticks or wheels can cause the pitch bend to jump erratically or simply remain unstable and non-zero when untouched by the user. This change would allow to ignore it completely in such cases as I see no way to do it currently in Bespoke.

aulerius:
This lowers the minimum range for the `pitchbendrange` setting in `midicontroller` to go as low as 0. It also changes the pitchbend calculation to avoid potential divison by zero.
This Closes #2015 in a pretty blunt way (`midioutput` and `vstplugin` are still limited to 1). Perhaps expanding `pitchbender` to allow _setting_ instead of only _adding_ would be smarter but I've wanted to try a very basic PR to solve an actual problem I'm having right now (and it's my first ever github PR!!! 🎉)

I'm also yet to build Bespoke myself so this is untested so far but it's such a tiny and contained change that it feels okay.

### DIFF
--- a/Source/MidiController.cpp
+++ b/Source/MidiController.cpp
@@ -475,7 +475,7 @@ void MidiController::OnMidiPitchBend(MidiPitchBend& pitchBend)
 
    int voiceIdx = -1;
 
-   float amount = (pitchBend.mValue - 8192.0f) * mPitchBendRange / 8192.0f;
+   float amount = (pitchBend.mValue - 8192.0f) * (mPitchBendRange / 8192.0f);
 
    if (mUseChannelAsVoice)
       voiceIdx = pitchBend.mChannel - 1;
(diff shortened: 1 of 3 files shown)


aulerius:
I've made a windows build including this change and works as expected on my end.

Please let me know if anything with this PR is unfitting to be merged, incomplete, or up for debate. Would like to know the details.

awwbees:
nothing controversial, I probably just want to take a look at `vstplugin` and `midioutput` to see if it makes sense to make this same change in parallel there, and I haven't gotten around to it yet

PR #2053, opened 2026-05-19
Introduce 'permutation' code for `binaryvalue`

merkys:
This PR adds one more code type for `binaryvalue`: 'permutation'. In this regime, binary numbers are ordered by the number of 1s in them. Thus the sequence of binary numbers consists of 0, all the binary numbers having one 1, then all binary numbers having two 1s and so on (see [OEIS sequence A047869](https://oeis.org/A047869)).

This code type might be useful to, for example, gradual increase of notes allowed through `notefilter` or tracks enabled in `drumsequencer`. Two previous modes ('byte' and 'gray') do not have this property.


merkys:
I cannot reproduce the macOS build failure (I have no macOS machines available). I tried replacing `c++` with `clang++` (I gather macOS uses that), but `clang++` 19.1.7 builds just fine. I would really appreciate any pointer regarding this failure.

PR #2059, opened 2026-05-26
feat: button in help area that checks for nightly updates

vrozkovec:
Using the most recent Nightly:

https://github.com/user-attachments/assets/e2ee6e42-5cac-497d-bf24-982e0bdfc7f3

Nightly update available - opens browser at https://github.com/BespokeSynth/BespokeSynth/releases/tag/Nightly:

https://github.com/user-attachments/assets/0af4c8e7-7ae1-47af-9a68-f87873613257


vrozkovec:
Fixed formatting, build now passes.

Label of the button is "check for new version".

PowerUser64:
Nice! Looks good to me.

Although, if I were using the nightly builds, I think I would be clicking this incessantly. I wonder if an "auto-check for updates" checkbox in settings with some kind of notification would be better? I don't like the idea of having bespoke ping github every time it starts though, so that's the plus of doing it this way.

vrozkovec:
We had a discussion on Discord, and it was decided to do it this way for now. 
The solution you propose also makes sense, but would require more changes, which @awwbees didn't want to do at the moment, and has a drawback of pinging some external service, as you've mentioned.

PowerUser64:
Just some thoughts for how to offer the "ping github on startup" method, we could have a user-defined interval to check for updates at (default to 7 to 14 days or something). We could save four keys in settings for it: `enable_startup_update_check` (bool), `update_check_interval_days` (int), `last_update_check` (unix timestamp), and `can_update` (bool, cached value of the last check, only run the check if it's `false`).

This way, we only ping github once per update, and very infrequently. If the user wanted the check to happen on every startup though, they could set the interval to zero and it would always happen. We could have a hidden minimum 1 hour though, just to prevent rate-limiting.

The button added in this PR could co-exist with it and should ignore the 1-hour limit, although I think it should stay only as part of the nightly builds. The release builds could use this setting too, but have a default interval of 180 days or something really long like that.

I'd be curious what @awwbees thinks about this.

awwbees:
I'm not interested in adding anything that feels like an automatic "phone home", it's low value and not worth the heebie-jeebies that it will give to privacy-sensitive users. the button in this PR is acceptable because it's very opt-in.

I'm also not interested in actively pushing users to always updating to the latest nightly build. even though they're typically always stable and savestate compatible, it's still technically an "unstable" build, and mostly suitable for the type of nerd that regularly checks the #github channel on discord.

PR #2060, opened 2026-05-30
add `djfilter` effect, for a dj mixer type of tilt filter

awwbees:


PowerUser64:
Wow, I never thought I'd be deleting prefabs, but here we are!

PR #2061, opened 2026-05-31
`tapelooper`: new simpler approach to retroactive loop recording

awwbees:


PowerUser64:
best thing since sliced bread!
sick ui, btw

PR #2064, opened 2026-06-01
tapelooper: add passthrough for when `mEnable` is false

PowerUser64:


awwbees:
a fix along these lines went in with #2066

PR #2065, opened 2026-06-01
tapelooper: remove unneeded "recording" checkbox

PowerUser64:
Feel free to close if this is intended, or if you have other ideas for where to take this!

The tapelooper module tracks its recording state in two places, through the "recording" checkbox and through state selector. This removes the checkbox in favor of just having the state selector.

before:
image

after:
image

### DIFF
--- a/Source/TapeLooper.cpp
+++ b/Source/TapeLooper.cpp
@@ -49,7 +49,6 @@ void TapeLooper::CreateUIControls()
    FLOATSLIDER(mLoopBeatsAgoSlider, "loop beats ago", &mLoopBeatsAgo, 0, 4 * 16);
    FLOATSLIDER(mDownbeatOffsetBeatsSlider, "downbeat offset", &mDownbeatOffsetBeats, 0, 4 * 16);
    UIBLOCK_NEWCOLUMN();
-   CHECKBOX(mRecordCheckbox, "record", &mRecording);
    BUTTON(mLoop1BarButton, "loop 1 bar");
    BUTTON(mLoop2BarsButton, "loop 2 bars");
    BUTTON(mLoop4BarsButton, "loop 4 bars");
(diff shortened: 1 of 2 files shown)


PowerUser64:
I think clang-format may have changed their defaults since version 14 (they are on v22). On my system, it formats it like this, but the ubuntu instance in CI disagrees:

```
Source/TapeLooper.h:70:80: error: code should be clang-formatted [-Wclang-format-violations]
   void IntSliderUpdated(IntSlider* slider, int oldVal, double time) override { }
                                                                               ^
Source/TapeLooper.h:71:83: error: code should be clang-formatted [-Wclang-format-violations]
   void RadioButtonUpdated(RadioButton* radio, int oldVal, double time) override { }
                                                                                  ^
```

Should I:
A) Use clang-format v14 locally
B) Update the clang-format file so v22 formats it like CI expects (not sure if the v14 format is forwards compatible with v22)
C) Update the CI pipeline to ubuntu 26.04 LTS to get new clang-format (and maybe option B as well so we don't need to format everything)
D) Something else?

This should probably be another PR.

PowerUser64:
Back to the main topic of this, I wonder if the selector is the wrong abstraction and we should be using `mEnable` to stop/start playback and have a separate `mRecording` checkbox. The disadvantage is that `mEnable` would also need to disable recording, but I don't think that's too big of a deal.

awwbees:
I originally just had the radio selector, but the checkbox isn't redundant, adds additional functionality: it tracks when you start recording, and sets the loop length accordingly when you disable the checkbox. this PR breaks that functionality. there's probably a better interface than the radio selector + recording checkbox, will continue to explore as I use the module.

for the clang-format issue, sounds like a separate PR. I'm not really up on the implications of updating it.

PR #2082, opened 2026-06-23
Accept OSC bundles in `osccontroller`

merkys:
This PR makes `osccontroller` aware of OSC bundles. The bundles are split into OSC messages and executed without respecting their simultaneity, just as done in `sampleplayer`.


merkys:
Thanks for merge!

PR #2089, opened 2026-07-11
New modules, visualizers, global palettes & search panel

n1m21n:
Hey! These are the modules and changes I've been building on my fork — putting them up as you suggested. Happy to split this into smaller PRs or adjust anything to fit the project's style.

New modules
- Wavetable — wavetable oscillator with per-oscillator filter + envelope, A/B oscillators, waveform previews, and unison.
- Tracker — sample step-sequencer (per-step sample/volume/decay/pitch/repeat, transport-synced, per-step + global randomize).
- MolderSampler ("molder") — unified sampler with 5 engines (classic / slice / repitch / granular / spectral), a global play-region with loop/ping-pong/reverse, onset + manual slicing, and live audio recording into the sampler.
- Formant filter effect.

New "Visualizers" module category (top bar):
- BlobViz — audio-reactive blob field.
- CubeViz — grid of parametric 3D shapes (cube, sphere, torus + several Paul-Bourke surfaces) with count, symmetry, randomize.
- PixelCloud — drop an image; renders as a glitchy ASCII/halftone field or a 3D point cloud.
- TextCloud — type text; extrudes into a rotatable 3D object with grain/blur/bloom/distortion + font selection.
- CheckerBox — animated optical-illusion checkerboard (grid / bend / spiral / tunnel) with chromatic aberration.

All visualizers are CPU immediate-mode (no shaders/FBO) with capped work per frame.

Shared changes
- 24 shared cosine colour palettes used across all visualizers (VizPalettes.h).
- A search panel docked to the UI layer (background sample indexing persisted to disk, browse + filter, draggable scrollbar).
- New kModuleCategory_Visualizer category wired into the module factory and title bar.

This is my first PR, so please let me know anything I should clean up. Thanks for the amazing project!


awwbees:
very cool stuff in here! I'm hitting some asserts whenever I try to spawn some of these new modules, which are there to help protect you from other problems. you're probably only running in a release build, so you're not encountering them. I'd recommend that you make a debug build and test to help catch these issues. some examples:
- wavetable: asserts because there are multiple sliders named "position" (this is an easy fix, you just need to give the sliders for oscillator b a different name, such as "position b")
- switcher: asserts because it's trying to add a transport listener before the module has been initialized

PR #2096, opened 2026-07-16
Convert all files to unix line endings

PowerUser64:
I was going to write this in a comment on #2095 but I figured it would be more useful if I just did it instead.

This:
- converts all files to unix (LF) line endings
	- changes files in `Source/`, `resource/`, and `autodoc/`
	- doesn't touch`libs/`
- tells `git blame` to ignore this conversion, as well as the one in #2095 (`.git-blame-ignore-revs`)
- asks editors to default to using LF line endings (with .editorconfig)

The second patch in this set is still good to have even if you don't care about the other existing files, since it will ignore the previous line-ending conversion in `git blame`. (Although it would require a change to not be applicable to just this PR.)

The third patch (the editorconfig change) pretty much requires the first one, since otherwise editors will auto-convert files to unix line endings, and you will effectively end up with the first patch over time

Feel free to close if you don't care about line endings that much.


awwbees:
yeah, I was surprised to find that the whole codebase was a mishmash of line endings, whoops. I was also thinking about converting them all over. the only thing that gives me pause is wondering if this is going to cause painful merge issues for all current PRs.

PowerUser64:
Closing since it doesn't seem like this will be mergeable soon. Maybe there's a way to do this using git that reduces conflicts.

It's fairly easy to recreate the bulk of this PR though, just run `dos2unix Source/*.{h,cpp,txt,i,in}`

PR #2105, opened 2026-08-18
add ability to set default value on sliders in code, which fixes `interface` not being able to reset sliders properly

awwbees:


PowerUser64:
Nice!

PR #2109, opened 2026-08-26
add python get functions for notesequencer

afentoe:
To be able to retrieve grid data from the NoteSequencer in Python, I have created get_* functions:
```
      n.get_row(step)
      n.get_pitch(step)
      n.get_velocity(step)
      n.get_length(step)
      n.get_step(step)
        example: row, pitch, velocity, length = n.get_step(3)
```


afentoe:
@awwbees  Ah yes, I totally missed the correct way to update the python ref docs. Thanks!

PR #2110, opened 2026-08-26
fix audiosyncer out-of-bounds memcopy when using 96kHz samplerate

afentoe:
This fixes crash when opening `audiosyncer` when using a sample rate of 96 kHz.

Fix 1:
RollingBuffer call to DrawAudioBuffer used bufferSize -1, modulo was disabled causing crash at 96 kHz.
Changed to actual buffer Size().
This prevents memcopy crash.

Fix 2:
AudioSyncer mBuffer.Draw could use invalid index ranges at high sample rates.
Added clamp to displaySamples  parameter.

### DIFF
--- a/Source/AudioSyncer.cpp
+++ b/Source/AudioSyncer.cpp
@@ -157,7 +157,13 @@ void AudioSyncer::DrawModule()
    }
 
    for (auto& bandBuffer : mBandBuffers)
+   {
+      if (displaySamples != -1) // clamp if not in static mode
+      {
+         displaySamples = std::min(displaySamples, bandBuffer->mBuffer.Size());
+      }
       bandBuffer->mBuffer.Draw(kBufferX, kBufferY, kBufferW, kBufferH, displaySamples, 0, 0, bandBuffer->mColor, false);
+   }
 
    ofPushStyle();
 
--- a/Source/AudioSyncer.cpp
+++ b/Source/AudioSyncer.cpp
@@ -157,13 +157,7 @@ void AudioSyncer::DrawModule()
    }
 
    for (auto& bandBuffer : mBandBuffers)
-   {
-      if (displaySamples != -1) // clamp if not in static mode
-      {
-         displaySamples = std::min(displaySamples, bandBuffer->mBuffer.Size());
-      }
       bandBuffer->mBuffer.Draw(kBufferX, kBufferY, kBufferW, kBufferH, displaySamples, 0, 0, bandBuffer->mColor, false);
-   }
 
    ofPushStyle();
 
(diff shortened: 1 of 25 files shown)


afentoe:
@awwbees 
Here is my userprefs.json
[userprefs.json](https://github.com/user-attachments/files/31504868/userprefs.json)

So using ASIO on Windows.: 96000 512
I added debug logging in several functions to output values of several vars, because on my dev/build pc, it also works fine, so could not debug.
If you want, I can try to recreate these logs to give some insights what the var values are, right before the crash.

I remember some logs with a float value of: 479999.0

AudioSyncer: displaySamples=-1 bufferSize=480000 static=1
RollingBuffer::Draw: length=-1 Size=480000 channel=0 delayOffset=0
DrawAudioBuffer(inner): start=0.0 end=480000.0 pos=-1.0 wrapFrom=-1 wrapTo=0 bufferSize=-1 width=800.0 height=200.0
DrawAudioBuffer(inner): length=479999.0


And over here https://github.com/BespokeSynth/BespokeSynth/blob/626ae10bebee7a0f3bac71ed1e778acfd4a21423/Source/SynthGlobals.cpp#L209
sampleIdx  could increase > 480000

awwbees:
still can't reproduce it, so let's try some back-and-forth debugging... without your other changes, at that section of SynthGlobals.cpp, at line 214, what if you add this?
```
if (sampleIdx > length - 1)
   sampleIdx = length - 1;
```
right above the `max = ` line? does that fix the crash for you? I'm still not totally sure what's going on, but this seems like a useful safety check anyway.

afentoe:
Reverted all changes, only applied your patch in SynthGlobals.cpp.

Result:
No more crashes, so that's good.

But new issue introduced:
When disabling `static waveform`  and decreasing the `display length`, I see periods of no waveform display.
These flat lines get longer when I decrease the `display length`.

Note: my initial commit did not have this issue.
When adding my commits to your new patch again, problem remains.
So def the new `sampleIdx = length - 1;` is the cause for these flatlines..

I will try to investigate this further.

image

afentoe:
Audiosyncer memcopy crash was fixed in the new PR https://github.com/BespokeSynth/BespokeSynth/pull/2114 by better calculating sampleIdx limits in DrawAudioBuffer.
But that PR introduced a flatline in the audiobuffer display when disabling `static waveform` and decreasing the `display length`.

Fix by removing these lines again:
```
if (sampleIdx > length - 1)
   sampleIdx = length - 1;
```

@awwbees not sure if you want to merge this PR, which now effectively only removes 2 lines. Perhaps rename this PR merge commit.

Or just remove the 2 lines from your main branch in a separate PR and close this one.

awwbees:
thank you for testing this!

PR #2111, opened 2026-08-28
fix zooming so it doesn't drift!

PowerUser64:
this has bothered me for years, finally took the time to math it out

here's how it works: https://www.desmos.com/calculator/nnoypjy1wk

video demo (before/after)

https://github.com/user-attachments/assets/a95af8b4-50e2-4850-9a5a-bbd667a0077b


awwbees:
nice one! never noticed this

PR #2113, opened 2026-09-01
filter out DC offset from delay feedback, to avoid the feedback result crawling away from zero offset

awwbees:


NSNS33:
Just a heads-up. This can lead to a change in sound compared to the older builds. A more robust solution would be a freely configurable feedback FX chain where users can decide on the DC offset according to their needs. Or at least a DC offset toggle.

